PR #2, opened 2023-03-31
Small fix so torch_harmonics.distributed gets installed

akshaysubr:
Was running into an issue with installing the package using `pip install .` or including this as a dependency in other repos using a `pyproject.toml`:

```
import torch_harmonics as harmonics
...
ModuleNotFoundError: No module named 'torch_harmonics.distributed'

ls /usr/local/lib/python3.8/dist-packages/torch_harmonics
__init__.py  __pycache__/  legendre.py  quadrature.py  sht.py  test_torch_harmonics.py
```

Using `find_packages()` fixes the issue for me by recursively adding package directories that have `__init__.py` files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
 nlat = 512
 nlon = 2*nlat
 batch_size = 32
-signal = torch.randn(batch_size, nlat, nlon)
+signal = torch.randn(batch_size, nlat, nlon).to(device)
 
 # transform data on an equiangular grid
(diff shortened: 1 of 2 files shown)


NickGeneva:
If you wanna stay explicit can also change it to `['torch_harmonics','torch_harmonics.distributed']` but find_packages is the way to go.

@bonevbs please have a quick look when you get the chance. Thanks.

bonevbs:
thanks @NickGeneva @akshaysubr , this should be addressed with `6aba153`. Let me know if issues persist.

PR #5, opened 2023-07-26
Fix __init__.py file to import from local "examples" dir.

m4e7:
Without this fix, importing torch_harmonics had run into a module import error when trying to execute:
> [37] `import examples`


m4e7:
@JeanKossaifi, this fixes the bug I mentioned last week.

PR #22, opened 2023-12-05
Tkurth/flexible sharding

azrael417:
This is a draft for the flexible sharding MR: new features are:

- sharding of tensors is more flexible, padding removed. This means that some GPU end up with smaller shards but the it removes the error prone padding logic users have to wrap their head around
- added all distributed tests to unites under the `torch_harmonics.distributed` folder
- vector SHT are now tested as well
- bumping up docker container version to 23.11
- bumping version number to 0.6.4

Remarks:
- there seem to be an initialization bug in `torch.irfft`. In my tests I am working around that but it was reported and we hope it gets fixed soon.


torch_harmonics/distributed/distributed_sht.py
```diff
@@ -556,42 +459,22 @@ def __init__(self, nlat, nlon, lmax=None, mmax=None, grid="lobatto", norm="ortho
...
-        londist = (self.nlon + self.comm_size_azimuth - 1) // self.comm_size_azimuth
-        self.nlonpad = londist * self.comm_size_azimuth - self.nlon
-
-        # frequency paddings
-        ldist = (self.lmax + self.comm_size_polar - 1) // self.comm_size_polar
-        self.lpad = ldist * self.comm_size_polar - self.lmax
-        mdist = (self.mmax + self.comm_size_azimuth - 1) // self.comm_size_azimuth
-        self.mpad = mdist * self.comm_size_azimuth - self.mmax
+        # compute splits 
+        self.lat_shapes = compute_split_shapes(self.nlat, self.comm_size_polar)
```
bonevbs, line 463: are those routines general enough to warrant a separate routine?

torch_harmonics/distributed/distributed_tests.py
bonevbs, line 1: woldn't it be better to have this in the tests folder? Of do you want to exclude it from the regular tests?

bonevbs (commented):
Thank you for making the changes. Overall, it looks already very good. I have added some minor comments

torch_harmonics/distributed/distributed_sht.py
azrael417: those were changed to be more flexible and it is absolutely crucial that those are in a separate routine. that way, one can always split the tensors consistently across ranks and we do not have to reimplement that for every layer with distributed tensors.

azrael417: this is basically a helper function which should be at the basis of a new distributed framework. it is used in some of the distributed primitives as well and ensures consistency in computing splits. if splits are not computed consistently, then one can easily produce NCCL hangs.

PR #30, opened 2024-02-01
Tkurth/distributed disco

azrael417:
this PR adds spatially distributed DISCO convolutions along with some unit tests.
Please note that the weight gradient reduction across polar and azimuth directions has to be performed outside this code either by using comm hooks or registered backward hooks. Not doing so will create diverging weights across ranks over time.

This MR also fixes some issues with the convolution GPU based unit tests.


bonevbs:
LGTM!

PR #32, opened 2024-03-13
Bbonev/cuda disco cleanup

bonevbs:
- Cleaned up conditional import based on whether the cuda extension is available
- removed all triton code
- cleaned up the setup system and added a flag to enforce the cuda build for containers etc.


azrael417:
lgtm

PR #34, opened 2024-03-25
Tkurth/cuda disco distributed fix

azrael417:
This MR fixes the distributed DISCO convs

### DIFF
--- a/torch_harmonics/distributed/distributed_convolution.py
+++ b/torch_harmonics/distributed/distributed_convolution.py
@@ -410,12 +410,7 @@ class DistributedDiscreteContinuousConvTransposeS2(DiscreteContinuousConv):
         x = gather_from_polar_region(x, -2, self.lat_in_shapes)
 
         # register allreduce for bwd pass
-        #x = copy_to_polar_region(x)
-
-        # DEBUG
-        #x = torch.sum(x, dim=2)
-        #out = scatter_to_polar_region(x, -2)
-        # DEBUG
+        x = copy_to_polar_region(x)
         
         if x.is_cuda and _cuda_extension_available:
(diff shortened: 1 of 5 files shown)


torch_harmonics/distributed/distributed_convolution.py
```diff
@@ -158,11 +152,12 @@ def _precompute_distributed_convolution_tensor_s2(in_shape, out_shape, kernel_sh
         out_vals.append(vals)
 
     # concatenate the indices and values
-    out_idx = torch.cat(out_idx, dim=-1)
-    out_vals = torch.cat(out_vals, dim=-1)
+    out_idx = torch.cat(out_idx, dim=-1).to(torch.long).contiguous()
```
bonevbs, line 155: why is the cast necessary here?

torch_harmonics/distributed/distributed_convolution.py
```diff
@@ -257,23 +243,18 @@ def get_psi(self):
         psi = torch.sparse_coo_tensor(self.psi_idx, self.psi_vals, size=(self.kernel_size, self.nlat_out_local, self.nlat_in_local * self.nlon_in)).coalesce()
         return psi
 
-    def forward(self, x: torch.Tensor, use_triton_kernel: bool = True) -> torch.Tensor:
+    def forward(self, x: torch.Tensor, use_cuda_kernel: bool = True) -> torch.Tensor:
```
bonevbs, line 246: I think this option should be removed. By default we'll use the cuda kernel if it is available

torch_harmonics/distributed/distributed_convolution.py
azrael417: It is safer since we pass it to type sensitive C++ code.

torch_harmonics/distributed/distributed_convolution.py
azrael417: fixed

PR #38, opened 2024-08-16
Tkurth/cuda disco

bonevbs:


Dockerfile
```diff
@@ -34,6 +34,10 @@ FROM nvcr.io/nvidia/pytorch:23.11-py3
 
 COPY . /workspace/torch_harmonics
 
+# we need this for tests
 RUN pip install parameterized
-RUN pip install /workspace/torch_harmonics
+
+# we need to remove old archs
```
azrael417, line 40: maybe we should write that arch < 7.0 does not support the atomics used in the disco kernel

azrael417 (approved):
Looks good to me.

Dockerfile
bonevbs: Good catch, I adapted the Readme file.

PR #40, opened 2024-08-19
adding reduce scatter and some small fixes

azrael417:
This MR adds reduce scatter in disco conv and its transpose as well as fixes some memory ordering edge cases in distributed transpose.


bonevbs:
LGTM, thanks

PR #41, opened 2024-08-21
Tkurth/distributed memory reduction

azrael417:
This rolls back the reduce scatter changes because of potential memory leaks which have to be investigated.


bonevbs:
LGTM

PR #42, opened 2024-08-21
Bbonev/disco fused quadrature

bonevbs:
Fused the multiplication of the quadrature weights into Psi, to slightly reduce the memory footprint of the DISCO convolution

### DIFF
--- a/torch_harmonics/convolution.py
+++ b/torch_harmonics/convolution.py
@@ -140,7 +140,7 @@ def _compute_support_vals_anisotropic(r: torch.Tensor, phi: torch.Tensor, nr: in
     return iidx, vals
 
 
-def _normalize_convolution_tensor_s2(psi_idx, psi_vals, in_shape, out_shape, kernel_shape, quad_weights, transpose_normalization=False, eps=1e-9):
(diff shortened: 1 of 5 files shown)


torch_harmonics/convolution.py
```diff
@@ -167,7 +167,11 @@ def _normalize_convolution_tensor_s2(psi_idx, psi_vals, in_shape, out_shape, ker
                 iidx = torch.argwhere((idx[0] == ik) & (idx[2] == ilat))
                 # normalize, while summing also over the input longitude dimension here as this is not available for the output
                 vnorm = torch.sum(psi_vals[iidx] * q[iidx])
-                psi_vals[iidx] = psi_vals[iidx] / (vnorm + eps)
+                if merge_quadrature:
+                    # the correction factor accounts for the difference in longitudinal grid points when the input vector is upscaled
+                    psi_vals[iidx] = psi_vals[iidx] * q[iidx] * nlon_in / nlon_out / (vnorm + eps)
```
azrael417, line 172: Why do we need the nlon_in / nlon_out ?

azrael417 (approved):
lgtm, but please check the comment

torch_harmonics/convolution.py
bonevbs: the quadrature weights that are used to normalize the transpose use the output resolution, however quadrature needs to run over the input. This has to be accounted for using this factor.

PR #43, opened 2024-08-23
Tkurth/distributed memory reduction

azrael417:
This MR fixes some issues with the fused reduce-scatter backward pass. Also, it ensures that activations should only be stored in requested AMP precision, and not in FP32 as was done accidentally previously.


torch_harmonics/distributed/distributed_convolution.py
```diff
@@ -418,8 +424,11 @@ def forward(self, x: torch.Tensor) -> torch.Tensor:
             x = distributed_transpose_azimuth.apply(x, (1, -1), self.lon_in_shapes)
 
         # gather input tensor and set up backward reduction hooks
-        x = gather_from_polar_region(x, -2, self.lat_in_shapes)
-        x = copy_to_polar_region(x)
+        if False:
```
bonevbs, line 427: same as above - why the `if False`

bonevbs (approved):
Looks good to me. Please double-check the if True/False statements whether we want to keep them as they are.

PR #48, opened 2024-08-28
fixing amp stuff again

azrael417:
This MR fixes some issues with type conversion in AMP and also adds some issues with non contiguous gradients which could impact performance in DDP


bonevbs (approved):
Looks good to me, thanks!

PR #52, opened 2024-09-09
fixing distributed sht

bonevbs:
The distributed SHT was assuming the channel/batching dimension to be dimension 3. This PR fixes that.

### DIFF
--- a/torch_harmonics/distributed/distributed_sht.py
+++ b/torch_harmonics/distributed/distributed_sht.py
@@ -108,9 +108,9 @@ class DistributedRealSHT(nn.Module):
         # combine quadrature weights with the legendre weights
         weights = torch.from_numpy(w)
         pct = _precompute_legpoly(self.mmax, self.lmax, tq, norm=self.norm, csphase=self.csphase)
-        pct = torch.from_numpy(pct)        
+        pct = torch.from_numpy(pct)
         weights = torch.einsum('mlk,k->mlk', pct, weights)
-        
+
         # split weights
         weights = split_tensor_along_dim(weights, dim=0, num_chunks=self.comm_size_azimuth)[self.comm_rank_azimuth]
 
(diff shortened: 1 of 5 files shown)


torch_harmonics/distributed/distributed_sht.py
```diff
@@ -234,12 +237,15 @@ def extra_repr(self):
 
     def forward(self, x: torch.Tensor):
 
+        if len(x.shape) < 3:
```
azrael417, line 240: I think you can use tensor.dims()

torch_harmonics/distributed/distributed_sht.py
```diff
@@ -360,14 +366,15 @@ def extra_repr(self):
...
+            raise ValueError(f"Expected tensor with at least 4 dimensions but got {len(x.shape)} instead")
+
         # we need to ensure that we can split the channels evenly
-        num_chans = x.shape[1]
+        num_chans = x.shape[-4]
 
         # h and w is split. First we make w local by transposing into channel dim
         if self.comm_size_azimuth > 1:
-            x = distributed_transpose_azimuth.apply(x, (1, -1), self.lon_shapes)
+            x = distributed_transpose_azimuth.apply(x, (-4, -1), self.lon_shapes)
```
azrael417, line 377: why is that -4 and not -3? Is -3 the vector index?

azrael417 (approved):
LGTM, just a few minor comments.

torch_harmonics/distributed/distributed_sht.py
bonevbs: right - I changed that

torch_harmonics/distributed/distributed_sht.py
bonevbs: correct - the vector SHT assumes that dimenion -3 is the vector dimension

PR #53, opened 2024-09-19
Bbonev/gradient test fix

bonevbs:
Fixing SHT gradient test to test SHT and ISHT individually, rather than the roundtrip. Also added analytical gradients to the gradient analysis notebook.


azrael417:
lgtm, feel free to merge

PR #54, opened 2024-10-01
safeguarding all custom CUDA and C++ routines via the _cuda_extension…

bonevbs:
- safeguardign all custom routines with the `_cuda_extension_available` flag
- removing torch as build dependency


azrael417 (approved):
Looks good to me. It is good to guard all the custom CUDA kernels.

PR #56, opened 2024-12-02
Adding Support for arbitrary point distributions on the sphere, as from DSE paper

LeviLingsch:
I have added several files and functions. 

`examples/train_fno_dse.py` provides the option to load two different models and train them via two slightly different functions. The models are `SFNODSEFp` and `SFNODSEVp` (SFNO using Discrete Spectral Evaluations with Fixed/Variable Points). "Fixed points" means sampling locations are fixed across all training data, while "Variable points" means that the sampling locations are different between the samples.

The spherical harmonics transformed is performed by the class `RealSHTDSE` for fixed points and `BatchedRealSHTDSE` for variable points. These are both found in `torch_harmonics/sht_dse.py`. These classes must be initialized with the set of points and number of degrees for the SHT. They contain functions to perform both the forward and backward transform. The backward transform is not an inverse. Actually, the backward transform will return a field with values that differ from the original field by some constant. This can be by about an order of magnitude. For the purposes of operator learning, this was actually not an issue, since some scaling of the SHT coefficients is learned. I never bothered to fix this, but if someone is hoping to use this package strictly for "inverse" SHT on an arbitrary point distribution, this might not work for them. 

The models for the SFNO on arbitrary points are found in `examples/sfno_dse/models`.


azrael417:
I am sorry for the very delayed reply, we totally dropped the ball on this. For unstructured meshes, those are a whole different beast and ideally we would write a mother library for this. Thanks for your work but currently we are not considering extending torch harmonics to unstructured mesh?

PR #59, opened 2024-12-12
changing defuault grid to equiangular in all SHT routines

bonevbs:
Changing the default grid throughout all SHT routines.


azrael417 (approved):
LGTM, go ahead

PR #63, opened 2025-01-16
Update README.md

Michael-T-McCann:
Without putting `signal` on `device`, we get `RuntimeError: Expected all tensors to be on the same device` when `sht` is called.


azrael417:
Hello Michael, thanks for catching and fixing this. You are correct.

bonevbs (approved):
thanks for catching this!

PR #66, opened 2025-01-21
Tkurth/torchification

azrael417:
This PR adds the following features:

- replacing most numpy pre-computations with torch pre-computations
- adding python function caching for the pre-computations


tests/test_distributed_resample.py
```diff
@@ -183,14 +183,14 @@ def _gather_helper_bwd(self, tensor, B, C, resampling_dist):
...
-            [128, 256, 64, 128, 32, 8, "equiangular", "equiangular", "bilinear-spherical", 1e-7],
+            [64, 128, 128, 256, 32, 8, "equiangular", "equiangular", "bilinear", 1e-7, False],
+            [128, 256, 64, 128, 32, 8, "equiangular", "equiangular", "bilinear", 1e-7, False],
+            [64, 128, 128, 256, 32, 8, "equiangular", "equiangular", "bilinear-spherical", 1e-7, False],
+            [128, 256, 64, 128, 32, 8, "equiangular", "equiangular", "bilinear-spherical", 1e-7, False],
         ]
     )
     def test_distributed_resampling(
-            self, nlat_in, nlon_in, nlat_out, nlon_out, batch_size, num_chan, grid_in, grid_out, mode, tol
+            self, nlat_in, nlon_in, nlat_out, nlon_out, batch_size, num_chan, grid_in, grid_out, mode, tol, verbose
```
bonevbs, line 193: does the verbose option exist in all tests now? Would be good to have it for consistency

torch_harmonics/convolution.py
```diff
@@ -258,7 +258,7 @@ def __init__(
         self,
         in_channels: int,
         out_channels: int,
-        kernel_shape: Union[int, List[int]],
+        kernel_shape: Union[int, Tuple[int]],
```
bonevbs, line 261: This should probably be `Union[int, Tuple[int], Tuple[int, int]]`

torch_harmonics/convolution.py
```diff
@@ -134,17 +135,18 @@ def _normalize_convolution_tensor_s2(
...
-    filter_basis,
-    grid_in="equiangular",
-    grid_out="equiangular",
-    theta_cutoff=0.01 * math.pi,
-    theta_eps = 1e-3,
-    transpose_normalization=False,
-    basis_norm_mode="mean",
-    merge_quadrature=False,
+    in_shape: Tuple[int],
+    out_shape: Tuple[int],
```
bonevbs, line 141: it should also work with Lists fyi.

azrael417: that is not possible unfortunately, since List is not a haschable datatype.

torch_harmonics/convolution.py
azrael417: fixed

torch_harmonics/distributed/distributed_resample.py
```diff
@@ -82,48 +81,51 @@ def __init__(
 
         # for upscaling the latitudes we will use interpolation
         self.lats_in, _ = _precompute_latitudes(nlat_in, grid=grid_in)
-        self.lons_in = np.linspace(0, 2 * math.pi, nlon_in, endpoint=False)
+        self.lons_in = torch.linspace(0, 2 * math.pi, nlon_in+1)[:-1]
```
bonevbs, line 84: maybe  it would be good to pack this into a `_precompute_longitudes` routine? I see this pattern repeat manytimes and this should improve readility

torch_harmonics/examples/pde_sphere.py
```diff
@@ -34,6 +34,7 @@
 import torch.nn as nn
 import torch_harmonics as harmonics
 
+import math
 import numpy as np
```
bonevbs, line 38: is it possible to remove the numpy import here? Ideally we should get rid of numpy for everything that gets imported with torch-harmonics

torch_harmonics/resample.py
```diff
@@ -67,48 +67,51 @@ def __init__(
...
         # lat_idx = np.where(self.lats_out < self.lats_in[0], 0, lat_idx)
 
         # compute the interpolation weights along the latitude
-        lat_weights = torch.from_numpy((self.lats_out - self.lats_in[lat_idx]) / np.diff(self.lats_in)[lat_idx]).float()
+        lat_weights = ((self.lats_out - self.lats_in[lat_idx]) / torch.diff(self.lats_in)[lat_idx]).to(torch.float32)
         lat_weights = lat_weights.unsqueeze(-1)
 
         # convert to tensor
-        lat_idx = torch.LongTensor(lat_idx)
+        #lat_idx = torch.LongTensor(lat_idx)
```
bonevbs, line 97: is this unnecessary here? If so we should delete it. Same for distributed case.

bonevbs (commented):
Looks already very good to me. Most comments are minor things. I would suggest adding some form of tests for the cache and the modified routines for the grid/quadrature precomputations to make sure they work as intended.

Please also update `pyprojerct.toml` and `Changelog.md`

torch_harmonics/distributed/distributed_resample.py
azrael417: done

torch_harmonics/resample.py
azrael417: I think it is already a long tensor

azrael417: done

tests/test_distributed_resample.py
azrael417: not all tests, just the ones which previously printed things

azrael417: I have added it to tests where it is relevant.

torch_harmonics/examples/pde_sphere.py
azrael417: his is in the plotting section, I think numpy is good there since matplotlib is naturally used with numpy

PR #70, opened 2025-04-24
setting imaginary parts of DCT and nyquist frequency to zero in IRSHT

azrael417:
This MR forces certain imaginary components to zero for inverse complex to real SHT. This is important since some backend implementations of irfft, used for transforming the m components, are not ensuring that on their own.


bonevbs:
Thank you for addressing this. Could you also have a look at the InverseVectorSHT? It should need this fix as well.

bonevbs (approved):
LGTM, thank you for addressing this

PR #77, opened 2025-06-11
Optimized CUDA kernels for S2 Attention (forward and backward)

rietmann-nv:
Forward kernel that uses a warp-per-output parallelization strategy (based on Mauro Bisson's implementation). Backward kernel uses the same strategy and was written by Andrea Paris and Max Rietmann.


bonevbs:
LGTM

PR #78, opened 2025-06-13
fixing attention perf test attempt 1

azrael417:
this MR fixes the attention perf test which cannot be run in the CI


bonevbs (approved):
LGTM

PR #80, opened 2025-06-16
Optimize bwd kernel: incremental qdot_max and alpha/integral/etc

rietmann-nv:
Leverage the same qdotk_max "trick" for the backward kernel. This avoids 1 loop and saves about 20% of performance.


azrael417:
LGTM

azrael417 (approved):
lgtm

PR #84, opened 2025-07-01
Small fix in metric computation

apaaris:
This PR fixes a small bug in the metrics computation


azrael417:
LGTM, thanks for catching that.

PR #86, opened 2025-07-03
torch.set_default_device fixes.

dallasfoster:
When setting `torch.set_default_device('cuda:0')` I ran into several errors because some devices are initialized `.from_numpy`, which automatically sets them on the cpu. This MR fixes those few places.

Furthermore, I ran into the following code that produces a segmentation fault:

```python
import torch_harmonics as th
import torch

torch.set_default_device('cuda:0')
th.convolution.DiscreteContinuousConvS2(
    in_channels = 1,
    out_channels = 1,
    in_shape = (180, 360),
    out_shape = (90, 180),
    kernel_shape = (3, 3),
    basis_type = "morlet",
    basis_norm_mode = "mean",
)
```

I included a fix, which ensures that the data going into `preprocess_psi` is located on the cpu.

### DIFF
--- a/torch_harmonics/convolution.py
+++ b/torch_harmonics/convolution.py
@@ -170,8 +170,7 @@ def _precompute_convolution_tensor_s2(
     kernel_size = filter_basis.kernel_size
 
     nlat_in, nlon_in = in_shape
-    nlat_out, nlon_out = out_shape
-
+    nlat_out, nlon_out = out_shape    
     # precompute input and output grids
     lats_in, win = _precompute_latitudes(nlat_in, grid=grid_in)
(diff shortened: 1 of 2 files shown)


azrael417:
Let me think about how we can fix that fully. We could move the pre computation back to torch. That way, we could also execute that on the device if we wanted.

azrael417:
Hello Dallas, can you please check the latest main branch and see if the error still persists?

bonevbs:
tagging @dallasfoster he might have missed this

dallasfoster:
The cleanup branch has resolved the issues I was having.

Resolved by #90

PR #87, opened 2025-07-04
Use 64-bit for pointer offsets

mauro-bis:
Updated pointer offset calculations to use 64-bit integers to prevent overflow with large batch or image sizes. Also, removed an unused parameter from an internal function.


azrael417 (approved):
LGTM, thanks for fixing this.

PR #88, opened 2025-07-07
removing some comments and nvtx annotations

azrael417:
This MR removes some debug comments and nvtx annotations inside the cuda code.


bonevbs (approved):
LGTM

PR #90, opened 2025-07-09
Tkurth/cleanup

azrael417:
Simplifying the precompute convolution tensor routine, re-using the single rank one for distributed as well. 
Re-writing parts of the code to assure that the routine can be called with device tenors as well.

### DIFF
--- a/torch_harmonics/convolution.py
+++ b/torch_harmonics/convolution.py
@@ -147,6 +147,7 @@ def _precompute_convolution_tensor_s2(
     transpose_normalization: Optional[bool]=False,
     basis_norm_mode: Optional[str]="mean",
     merge_quadrature: Optional[bool]=False,
+    support_only: Optional[bool]=False,
 ):
     """
     Precomputes the rotated filters at positions $R^{-1}_j \omega_i = R^{-1}_j R_i \nu = Y(-\theta_j)Z(\phi_i - \phi_j)Y(\theta_j)\nu$.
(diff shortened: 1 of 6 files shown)


torch_harmonics/filter_basis.py
```diff
@@ -197,25 +202,29 @@ def _compute_support_vals_anisotropic(self, r: torch.Tensor, phi: torch.Tensor,
...
-            dist_rn = (rn[iidx[:, 1], iidx[:, 2]] - ir[iidx[:, 0], 0, 0]).abs()
-            dist_phin = _circle_dist(phin[iidx[:, 1], iidx[:, 2]], iphi[iidx[:, 0], 0, 0])
-            # compute the value of the basis functions
-            vals = cond_r[iidx[:, 0], iidx[:, 1], iidx[:, 2]] * (1 - dist_r / dr)
-            vals *= cond_phi[iidx[:, 0], iidx[:, 1], iidx[:, 2]] * (1 - dist_phi / dphi)
-            valsn = cond_rn[iidx[:, 0], iidx[:, 1], iidx[:, 2]] * (1 - dist_rn / dr)
-            valsn *= cond_phin[iidx[:, 0], iidx[:, 1], iidx[:, 2]] * (1 - dist_phin / dphi)
-            vals += valsn
+
+            if not support_only:
```
bonevbs, line 206: same here - see below. I would maybe define a separate function to compute the support and use that. I don't think this belows in the basis

torch_harmonics/convolution.py
```diff
@@ -218,35 +230,37 @@ def _precompute_convolution_tensor_s2(
...
+        iidx, vals = filter_basis.compute_support_vals(theta, phi, r_cutoff=theta_cutoff_eff, support_only=support_only)
 
         # add the output latitude and reshape such that psi has dimensions kernel_shape x nlat_out x (nlat_in*nlon_in)
         idx = torch.stack([iidx[:, 0], t * torch.ones_like(iidx[:, 0]), iidx[:, 1] * nlon_in + iidx[:, 2]], dim=0)
 
-        # append indices and values to the COO datastructure
+        # append indices and values to the COO datastructure, compute row offsets
         out_idx.append(idx)
         out_vals.append(vals)
+        out_roff[t + 1] = out_roff[t] + iidx.shape[0]
```
bonevbs, line 241: what is this for?

bonevbs: ah nevermind - just saw it. Did you check that the computation of roff is correct and consistent with what we had before?

PR #91, opened 2025-07-16
Attention Backward improvement

azrael417:
This MR includes Mauros latest optimizations for the backward path of the attention layer. I have refactored it a bit to make it more flexible if we want to introduce reduced precision support in the future. It is tested on the segmentation problem. I compared to a bunch of previous runs, loss curves look similar but slightly worse, but that might be by chance. I will do another run to see. Otherwise, this MR passes all the tests we have.

### DIFF
--- a/torch_harmonics/csrc/attention/attention_utils.cuh
+++ b/torch_harmonics/csrc/attention/attention_utils.cuh
@@ -262,13 +262,13 @@ void  permute_to0231_k(const int nchn,
     return;
 }
 
-template<int TRANSP_WARPS_X_TILE_SIZE, typename VAL_T>
+template<int WARPS_X_TILE, typename VAL_T>
 void launch_permute_to0231(at::Tensor src, at::Tensor dst){
     dim3 block;
     dim3 grid;
 
     block.x = WARP_SIZE;
-    block.y = TRANSP_WARPS_X_TILE_SIZE;
+    block.y = WARPS_X_TILE;
     grid.x = DIV_UP(src.size(1), block.x);
     grid.y = DIV_UP(src.size(3), block.x);
     grid.z = src.size(2)*src.size(0);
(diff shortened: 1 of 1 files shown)


torch_harmonics/csrc/attention/attention_fwd_cuda.cu
```diff
@@ -585,11 +374,13 @@ static void s2_attention_dipatch(int batch_size,
                                  at::Tensor row_off,
                                  at::Tensor col_idx,
                                  at::Tensor quad_weights,
-                                 at::Tensor yP,
-                                 cudaStream_t stream) {
+                                 at::Tensor yP) {
 
     static_assert(0 == (MAX_LOCAL_ARR_LEN & (MAX_LOCAL_ARR_LEN-1)));
 
+    // get stream
```
bonevbs, line 381: Don't we want to be able to specify the tream from within PyTorch?

azrael417: that is the way to do it. when using a stream context manager, the current stream will be the one from the context manager, otherwise the default stream for forward or backward, depending on when that routine is called.

mauro-bis (dismissed):
Looks fine to me.

torch_harmonics/_neighborhood_attention.py
```diff
@@ -520,6 +520,16 @@ def backward(ctx, grad_output):
         B, _, H, W  = grad_output.shape
         grad_output = grad_output.reshape(B*nh, -1, H, W)
 
+        # save type and convert to float32
+        kw_dtype = kw.dtype
+        vw_dtype = vw.dtype
+        qw_dtype = qw.dtype
+
+        kw = kw.to(torch.float32).contiguous()
```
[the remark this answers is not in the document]
azrael417: Looking at the kernel, I think float64 is not supported so this conversion is the right thing to do.

PR #93, opened 2025-07-21
Tkurth/device fixes

azrael417:
This enables instantiation of the set and disco layers on the device (currently not recommended due to potential precision issues, but enabled)


bonevbs (approved):
LGTM

PR #100, opened 2025-08-21
fixing docstrings in examples

azrael417:
This MR fixes some docstrings in the examples. Especially argument naming and ordering.


alxmrs:
Thanks! These new docstrings look good to me. I appreciate you updating them. 

One high level note of feedback: one thing that could help keep these up-to-date is to add a warnings-only CI check to run doctests: https://docs.python.org/3/library/doctest.html. Totally optional, but may serve as a useful reminder.

PR #101, opened 2025-08-27
Fixing attention layer for different numbers of embed dims

azrael417:
This fixes the PyTorch, OpenMP and CUDA implementations of the attention layer. Previously, the out_channels option did not work if out_channels was different from in_channels. Now the kernel exhibits the right behavior.


bonevbs (dismissed):
LGTM

bonevbs (approved):

PR #129, opened 2025-10-08
adding anisotropic morlet tests

azrael417:
Adding more tests for anisotropic morlet in disco


apaaris (approved):
Seems good to me! LGTM

PR #139, opened 2025-11-18
fix(swe): reduce prediction timestep to 30 minutes for stability

rietmann-nv:
- Change dt from 3600s to 1800s to address instability in attention models
- Add comments explaining stability issues with dt=3600s during autoregressive rollout
- Note that LSNO and SFNO remain stable at 3600s


bonevbs (approved):
LGTM

PR #140, opened 2025-11-24
Tkurth/distributed sht updates

azrael417:
This adds a small update to distributed SHT to reproduce the exact same behavior as serial SHT


torch_harmonics/distributed/distributed_sht.py
```diff
@@ -164,7 +165,12 @@ def forward(self, x: torch.Tensor):
         x = torch.view_as_real(x)
 
         # contraction
-        xs = torch.einsum('...kmr,mlk->...lmr', x, self.weights.to(x.dtype)).contiguous()
+        out_shape = list(x.size())
```
bonevbs: this only modifies the RealSHT. We should also modify InverseSHT and VectorSHTs accordingly

azrael417: Those already do the same for distributed and serial.

azrael417: This is resolved in inverse sht as well

PR #157, opened 2026-03-12
Introduce benchmark framework using CUDA events

mcgibbon:
This PR adds timing for the SHT and for the torch implementation of DISCO convolution through a new benchmarking framework, run through `python -m torch_harmonics.benchmark`.

This is largely taken from the implementation we used/I authored in https://github.com/ai2cm/ace

### DIFF
--- a/torch_harmonics/benchmark/benchmark.py
+++ b/torch_harmonics/benchmark/benchmark.py
@@ -39,10 +39,15 @@ class BenchmarkResult:
 class BenchmarkABC(abc.ABC):
     @classmethod
     @abc.abstractmethod
-    def new(cls: type[Self]) -> Self:
+    def new(cls: type[Self], batch_size_factor: float = 1.0) -> Self:
         """
         Initialize any state needed for the benchmark.
         This will be called once before the benchmark is run.
+
+        Args:
+            batch_size_factor: Combined hardware and user-requested scale factor
+                to apply to the base batch size. Each benchmark defines its own
+                base batch size and multiplies it by this factor.
         """
         pass
 
@@ -58,9 +63,9 @@ class BenchmarkABC(abc.ABC):
             torch.cuda.synchronize()
 
     @classmethod
-    def run_benchmark(cls, iters=10, warmup=1) -> BenchmarkResult:
+    def run_benchmark(cls, iters=10, warmup=1, batch_size_factor: float = 1.0) -> BenchmarkResult:
         null_timer = NullTimer()
-        benchmark = cls.new()
+        benchmark = cls.new(batch_size_factor=batch_size_factor)
         for _ in range(warmup):
             benchmark.run_instance(null_timer)
         timer = cls._make_timer()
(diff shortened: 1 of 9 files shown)


torch_harmonics/benchmark/benchmark.py
```diff
@@ -0,0 +1,109 @@
...
+        cls._sync()
+        cpu_timer.record_end()
+        return BenchmarkResult(
+            device=str(get_device()),
+            timer=timer.result,
+            cpu_time=cpu_timer.elapsed_time_ms(),
+        )
+
+    @abc.abstractmethod
+    def run_instance(self: Self, timer: Timer) -> TensorDict:
```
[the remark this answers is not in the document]
azrael417: I am not sure I follow. Yes, you have overhead when launching a single layer/kernel but you can compare that kernel to other variants since the overhead is similar. For the true kernel performance, one can also collect a profile and look at the raw timings in the profile.

torch_harmonics/benchmark/run.py
```diff
@@ -0,0 +1,156 @@
...
+    )
+    parser.add_argument(
+        "--device",
+        type=str,
+        default="cuda" if torch.cuda.is_available() else "cpu",
+        help="Device to run benchmarks on, e.g. 'cpu', 'cuda', 'cuda:1'. "
+             "Defaults to 'cuda' if available, otherwise 'cpu'.",
+    )
+    parser.add_argument(
+        "--batch-size-factors",
```
azrael417, line 136: Can we make this explicit batch size override instead of a factor?

azrael417 (commented):
I think having a batch size override is very important for quick kernel profiling and optimization iterations. We never run batch sizes > 1 and I do not want to have to edit a config and do PR when I change some benchmark settings. Since this is no competition benchmark and just for quick testing, I do not see a problem with having some flexibility here. One can write all the benchmark parameters into the benchmark file as well so that the results are comparable.

torch_harmonics/benchmark/run.py
mcgibbon: Alright, I will add this when I get back in a couple weeks.

azrael417:
Hello Jeremy, 

sorry for the delay but we had to update a lot of things in the repo and also release 0.9.1. Now we can have a look at the benchmark. Here are some remarks from my side:

1) the benchmark is not python 3.9 compatible and it will break there. Please replace typing.Self with the 3.9-safe pattern (string return annotations / TypeVar), and drop runtime | unions

2) add the missing license headers

3) Run the pre-commit hooks we have added to the repo for formatting and cleanup.

The style is very different from the plain style we are using in the rest of the code. While I am not a friend of heavy OOP in Python it is OK as long as it resides in its own subsystem. Generally, having a benchmark is great, thanks for adding this.

If you have time, can you have a look at my remarks and let me know what you think.

Best regards
Thorsten

mcgibbon:
Is this superseded by #225 or would you still like it updated?

azrael417:
Hello Jeremy, sorry for the late reply. Yes, I think we can close this MR. We decided to restructure the benchmark to be less OOP and uncouple the sizes of the benchmarks from the GPU model. But thank you very much for the effort and hopefully you find it useful.

PR #158, opened 2026-03-12
Add FFT-based DISCO S2 contraction for faster spherical convolutions

mcgibbon:
## Summary
  - Replace the loop-and-roll longitude correlation in DISCO S2 convolutions with FFT-based circular cross-correlation, enabled via use_fft_contraction=True on DiscreteContinuousConvS2 and DiscreteContinuousConvTransposeS2
  - Add a benchmark framework (torch_harmonics.benchmark) with a registry pattern, CUDA event timing, hardware-dependent batch size scaling, and a CLI runner (python -m torch_harmonics.benchmark)
  - Register benchmarks for RealSHT, InverseRealSHT, and DISCO convolutions (both torch sparse and FFT paths)

## Benchmark results (Tesla T4)

```
  ┌─────────────────────────────────┬───────────────┐
  │            Benchmark            │ Avg time (ms) │
  ├─────────────────────────────────┼───────────────┤
  │ disco_conv_s2_torch_1deg (B=16) │ ~97           │
  ├─────────────────────────────────┼───────────────┤
  │ disco_conv_s2_cuda_1deg (B=64)  │ ~24           │
  ├─────────────────────────────────┼───────────────┤
  │ disco_conv_s2_fft_1deg (B=64)   │ ~29           │
  └─────────────────────────────────┴───────────────┘
```

These results correspond to a 13.4x speed-up of the per-sample time. At equal batch size (B=16), the FFT path is ~10x faster than the torch sparse path (lower because the T4 has lower utilization for the FFT path). With this optimization, the torch-only timings are in the ballpark of the optimized CUDA timings, at least for this T4 GPU at 1-degree resolution.

## Test plan

  - pytest tests/test_fft_contraction.py — verifies FFT and torch paths produce assert_close-compatible results for forward, transpose, and gradients across multiple grid/basis/stride configurations
  - python -m torch_harmonics.benchmark — runs all registered benchmarks and saves JSON results

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 __pycache__
 *.so
 checkpoints
+*benchmark_results
--- a/tests/test_convolution.py
+++ b/tests/test_convolution.py
@@ -223,6 +223,25 @@ class TestDiscreteContinuousConvolution(unittest.TestCase):
             # transpose convolution
             [8, 4, 2, (16, 32), (16, 32), (3), "piecewise linear", "mean", "equiangular", "equiangular", torch.bfloat16, True, 5e-2, 5e-2],
             [8, 4, 2, (12, 24), (24, 48), (2, 2), "morlet", "mean", "equiangular", "equiangular", torch.bfloat16, True, 5e-2, 5e-2],
+            # FFT contraction tests (fp32 only)
+            # regular convolution
+            [8, 4, 2, (16, 32), (16, 32), (3), "piecewise linear", "mean", "equiangular", "equiangular", torch.float32, False, 1e-4, 1e-4, True],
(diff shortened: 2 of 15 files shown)


azrael417 (commented):
Thanks for this MR, especially the benchmarking. I left some comments. In particular, I am not sure how useful the FFT accelerated DISCO is since the psi tensor can become very big relatively quickly when not stored as sparse tensor. Also, please make sure there is feature equality between the distributed and non-distributed routines.

bonevbs:
Thank you for the pull-request @mcgibbon . We are closing this for the time being given that CPU execution speed  of DISCO has been addressed by #192. I'll keep the branch open as a future reference if we decide to implement this strategy for accelerating the CUDA kernels using FFTs near the polar caps as you suggested.

PR #164, opened 2026-04-07
Pre-Build Wheels

azrael417:
this is the rebased wheel branch written by Boris onto the current main branch.


.github/workflows/deploy_kitmaker.yml
bonevbs, line 1: is this script just for dry-runs?

build_wheels.py
bonevbs, line 1: isn't this handled already by the github integration?

Changelog.md
```diff
@@ -2,6 +2,32 @@
...
+* Added double precision support for DISCO convolution
+* Fixed Schmidt normalization for derivatives of associated Legendre polynomials
+* Fixed up/downsampling in attention layers when input and output shapes differ
+* Fixed `GaussianRandomFieldS2` to use `isht.lmax`/`isht.mmax` for compatibility with revised truncation logic
+* Distributed module: added shape verification for transpose and gather operations, controllable via `TORCH_HARMONICS_DISTRIBUTED_DEBUG`
+* Distributed module: fixed `finalize()` bug where process group was not properly destroyed
+* Query functions `torch_harmonics.disco.optimized_kernels_is_available` and `torch_harmonics.attention.optimized_kernels_is_available` for checking optimized layer availability
+* Quadrature helper functions `precompute_latitudes` and `precompute_longitudes` are now public API
+
+**Tests**
```
bonevbs: please keep it to a simple list per release, even if we have a lot of bulletpoints for 0.9.0

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = [ "setuptools", "setuptools-scm", "torch>=2.4.0", "numpy>=1.22.4"]
+requires = [ "setuptools", "setuptools-scm"]
```
bonevbs, line 2: is this right? technically, running the install requires torch right?

pyproject.toml
```diff
@@ -37,13 +37,13 @@ classifiers = [
...
-    "torch>=2.4.0",
+    "torch>=2.6.0",
     "numpy>=1.22.4",
 ]
 
 
-[tool.setuptools.dynamic]
-version = {attr = "torch_harmonics.__version__"}
+[tool.setuptools_scm]
+fallback_version = "0.9.0"
```
bonevbs, line 46: why doesn;t this get its version from the `torch_harmonics.__version__` field anymore?

bonevbs (commented):
please have a look at my comments.

build_wheels.py
azrael417: what do you mean? Here I explicitly exclude files so that they accidentally get committed. If we keep those strictly separate, we can remove these lines.

pyproject.toml
azrael417: This is a technicality to fix a circular dependency issue: basically we need to query the build system to find out what torch version we need. This is why the older builds failed in the original wheels branch. This fixes this issue. it will nevertheless install the right torch, it is just that it needs to be deleted from here.

pyproject.toml
azrael417: This is the fallback version. The version string needs to be overridable by SCM, that is how it works. Also, this has to be a static string. What we can do is to augment the scripts to dynamically populate this via (Claude suggestion):                     
                                                                                                                                                                                         
  But there's a cleaner fix: the CI workflow is already reading fallback_version from pyproject.toml to construct SETUPTOOLS_SCM_PRETEND_VERSION:                                        
                                                                                                                                                                                         
  BASE_VERSION=$(python -c "import tomllib; print(tomllib.load(open('pyproject.toml', 'rb'))['tool']['setuptools_scm']['fallback_version'])")                                            
                                                                                                                                                                                         
  You could just flip that — make __init__.py the single source of truth and have the workflow read from it instead:                                                                     
                                                                                                                                                                                         
  BASE_VERSION=$(python -c "import re; print(re.search(r'__version__\s*=\s*[\"\'](.*?)[\"\']', open('torch_harmonics/__init__.py').read()).group(1))")                                   
                                                                                                                                                                                         
  Then drop fallback_version from pyproject.toml (or keep it as a no-op backup). Right now "0.9.0" is duplicated in both __init__.py:32 and pyproject.toml:46, so one of them will       
  inevitably get out of sync when you bump the version.   

Shall we implement it that way?

.github/workflows/deploy_kitmaker.yml
azrael417: This is for future proofing once they enable pushing from external sources. Might never happen though, we can remove it from the repo if you want?

bonevbs: yes let us remove for now, I want to keep things minimal

build_wheels.py
bonevbs: I mean the build_wheels.py script seems to be redundant with the gihub CI which builds the wheels. Do we need the extra script? I think it would be good to have a central place where the wheels are built

pyproject.toml
bonevbs: in this case I prefer to hardcode it in pyproject.toml rather than in init.py. Is that possiible?

bonevbs (approved):
LGTM

apaaris (approved):
This looks good, thanks!

PR #170, opened 2026-04-17
adding fused spectral contraction backend with bf16/fp16 support

Anpandoh:
Adds a compiled spectral extension with fused contraction paths, cuBLAS-based complex GEMM decomposition, prepacked inference weights, and parity/benchmark coverage to validate numerical stability and performance gains.


azrael417:
Hello Anpandoh, thanks for reaching out. So this MR basically re-implements the complex contraction inside the spectral conv. I am always welcoming to performance improvements but I have to say that this here adds a lot of LOC to achieve something which can be done in a single line. The spectral contraction is not a bottleneck in most application is have seen so far and this is why it is not high on our list to optimize it. Note that with a bf16/fp16 contraction, you would also add fp32 <> bf16/fp16 conversion overhead because the SHT which precedes and succeeds the contraction commonly has to be done int FP32. Therefore, I do not see how improving this kernel can be useful.

Anpandoh:
Hi sorry, this was actually never was supposed to be pushed to the main repo, and was just created for testing purposes, will close now! Thanks!

PR #171, opened 2026-04-22
Fixing Downsampling in Attention

azrael417:
This mode properly handles downsampling in the NeighborhoodAttentionS2 layer. It also adds better error capturing and reporting for disallowed nlon_in/nlon_out combinations.


bonevbs (commented):
Looks mostly good to me. If the cross-attention or downsampling attention was tested and works properly, we can merge.

Please also update the changelog/version

apaaris (approved):
Thank you!

PR #172, opened 2026-04-22
Refactoring Custom Autograd

azrael417:
This MR moves the weight updates out of the backward pass for attention, because it can be solely handled by pytorch and is less error prone. It also greatly simplifies all kernels and they focus on the main custom part.

Also, we added qk-norm support to all attention layers which is important for stability in latent space rollouts.


apaaris (approved):
Thanks, lgtm

apaaris (approved):
Thank you

PR #173, opened 2026-04-23
some fixes to routines to avoid silent corruption and failures

azrael417:
some small fixes and cleanups.

### DIFF
--- a/tests/test_convolution.py
+++ b/tests/test_convolution.py
@@ -462,9 +462,6 @@ class TestDiscreteContinuousConvolution(unittest.TestCase):
         is_amp = dtype in (torch.float16, torch.bfloat16)
         module_dtype = torch.float32 if is_amp else dtype
 
-        if is_amp and self.device.type == "cpu":
-            raise unittest.SkipTest("skipping test because CPU does not support AMP autocast for this dtype")
-
         # set seed
         set_seed(333)
 
@@ -652,9 +649,6 @@ class TestDiscreteContinuousConvolution(unittest.TestCase):
 
         if (self.device.type == "cuda") and (not cuda_kernels_is_available()):
             raise unittest.SkipTest("skipping test because CUDA kernels are not available")
-
-        if is_amp and self.device.type == "cpu":
-            raise unittest.SkipTest("skipping test because CPU does not support AMP autocast for this dtype")
 
         # set seed
         set_seed(333)
--- a/tests/test_convolution.py
+++ b/tests/test_convolution.py
@@ -55,6 +55,12 @@ _devices = [(torch.device("cpu"),)]
 if torch.cuda.is_available():
     _devices.append((torch.device("cuda"),))
 
+# PyTorch CPU autocast doesn't support fp16/bf16 autograd for our ops, so AMP parameter
+# rows are only meaningful on CUDA. Gate them at collection time to avoid skip noise on
+# CPU-only runs (the in-test SkipTest still handles the CPU class variant when CUDA is
+# available, since parameterized_class also generates that variant).
+_AMP_SUPPORTED = torch.cuda.is_available()
+
 # perf thresholds
 # CPU results normalized to 16 OpenMP threads,
 # GPU results normalized to V100 16 GB GPU
@@ -89,6 +95,29 @@ def _normalize_convolution_tensor_dense(
     else:
         n_olat = nlat_out
 
+    # "none" skips bias/scale computation entirely; still mirror the sparse layout by rolling
+    # the r=0 slice into the other r slices (the sparse impl stores only r=0 and rolls at
+    # contraction time), then apply optional quadrature merging
+    if basis_norm_mode == "none":
+        pscale = nlon_in // nlon_out
+        for ik in range(kernel_size):
+            for ilat in range(n_olat):
+                if transpose_normalization:
+                    for r in range(1, nlon_out):
+                        psi[ik, :, r, ilat, :] = torch.roll(psi[ik, :, 0, ilat, :], r * pscale, dims=-1)
+                else:
+                    for r in range(1, nlon_out):
+                        psi[ik, ilat, r, :, :] = torch.roll(psi[ik, ilat, 0, :, :], r * pscale, dims=-1)
+
+        if transpose_normalization:
+            if merge_quadrature:
+                psi = quad_weights.reshape(1, -1, 1, 1, 1) * psi / correction_factor
+        else:
+            if merge_quadrature:
+                psi = quad_weights.reshape(1, 1, 1, -1, 1) * psi
+
+        return psi
+
     bias_arr = torch.zeros(kernel_size, n_olat, dtype=psi.dtype, device=psi.device)
     scale_arr = torch.zeros(kernel_size, n_olat, dtype=psi.dtype, device=psi.device)
     support_arr = torch.zeros(kernel_size, n_olat, dtype=psi.dtype, device=psi.device)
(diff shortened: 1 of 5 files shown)


tests/test_convolution.py
```diff
@@ -55,6 +55,12 @@
 if torch.cuda.is_available():
     _devices.append((torch.device("cuda"),))
 
+# PyTorch CPU autocast doesn't support fp16/bf16 autograd for our ops, so AMP parameter
+# rows are only meaningful on CUDA. Gate them at collection time to avoid skip noise on
+# CPU-only runs (the in-test SkipTest still handles the CPU class variant when CUDA is
+# available, since parameterized_class also generates that variant).
+_AMP_SUPPORTED = torch.cuda.is_available()
```
bonevbs: I think this should not check whether CUDA is available but rather on what device the test code is running

tests/test_convolution.py
```diff
@@ -89,6 +95,29 @@ def _normalize_convolution_tensor_dense(
...
+                        psi[ik, ilat, r, :, :] = torch.roll(psi[ik, ilat, 0, :, :], r * pscale, dims=-1)
+
+        if transpose_normalization:
+            if merge_quadrature:
+                psi = quad_weights.reshape(1, -1, 1, 1, 1) * psi / correction_factor
+        else:
+            if merge_quadrature:
+                psi = quad_weights.reshape(1, 1, 1, -1, 1) * psi
+
+        return psi
```
bonevbs: whpy do we need this early exiting business? I don't like that this adds further code duplication. I am happy wityh duplicating the computation in testing if this means less code and therefore better maintainability

tests/test_convolution.py
azrael417: that would need a refactor of the tests, with a different base test class for CPU and GPU. can be done but maybe in a separate MR

tests/test_convolution.py
```diff
@@ -120,24 +149,30 @@ def _normalize_convolution_tensor_dense(
...
 
+    # precompute the per-ik mean for "mean" mode so we don't rely on Python function-scope
+    # reuse of b/s across ilat iterations inside the loop below
+    if basis_norm_mode == "mean":
+        bias_per_ik = bias_arr.mean(dim=1)
+        scale_per_ik = scale_arr.mean(dim=1)
+
+    # precompute the "geometric" scalar once; it's ik/ilat-independent
+    if basis_norm_mode == "geometric":
+        geometric_scale = (1.0 - math.cos(theta_cutoff)) / 2.0 / 2.0
```
bonevbs, line 131: This was fine as is. better to put the code where it is needed

torch_harmonics/disco/convolution.py
```diff
@@ -136,6 +146,14 @@ def _normalize_convolution_tensor_s2(
     # get the quadrature weights
     q = quad_weights[ilat_in].reshape(-1)
 
+    # "none" skips bias/scale computation entirely; only quadrature merging matters
```
bonevbs: once again - why is this necessary?

torch_harmonics/disco/convolution.py
```diff
@@ -169,18 +197,14 @@ def _normalize_convolution_tensor_s2(
...
             elif basis_norm_mode == "support":
                 b = 0.0
                 s = support[ik, ilat]
-            elif basis_norm_mode == "none":
-                b = 0.0
-                s = 1.0
             elif basis_norm_mode == "geometric":
                 b = 0.0
-                s = (1.0 - math.cos(theta_cutoff)) / 2.0 / 2.0
+                s = geometric_scale
```
bonevbs, line 199: same as above

torch_harmonics/distributed/primitives.py
```diff
@@ -128,35 +128,45 @@ class _DistributeTransposeAzimuth(torch.autograd.Function):
     @custom_fwd(device_type="cuda")
     def forward(ctx, x, dims, dim1_split_sizes):
 
+        # short-circuit when the azimuth group isn't initialized (e.g., single-rank runs)
+        if not is_distributed_azimuth():
+            return x
```
bonevbs, line 133: I don't understand why this is correct

torch_harmonics/filter_basis.py
```diff
@@ -489,23 +520,29 @@ def kernel_size(self) -> int:
...
+
+        nr and nphi are accepted for signature compatibility with the base class and are
+        unused here — the analytic formula doesn't need a discretization grid.
 
-        Radial: integral_0^1 J_m(alpha r)^2 r dr = J_{m+1}(alpha)^2 / 2.
+        Radial: integral_0^R J_m(alpha r/R)^2 r dr = R^2 * J_{m+1}(alpha)^2 / 2.
         Angular: integral_0^{2pi} cos^2(m phi) dphi = 2pi (m=0) or pi (m>0).
         """
-        from scipy.special import jn as scipy_jn
+        del nr, nphi
```
bonevbs: why do we delete them here?

torch_harmonics/filter_basis.py
```diff
@@ -38,6 +38,17 @@
 
 from torch_harmonics.cache import lru_cache
 
+# scipy is only required for FourierBesselFilterBasis, which needs Bessel functions.
+# Import lazily at module load so plain users who don't need that basis aren't forced
+# to install scipy; availability is checked in FourierBesselFilterBasis.__init__.
+try:
+    import numpy as np
```
bonevbs: Shouldn't import numpy be outside of this block

torch_harmonics/filter_basis.py
```diff
@@ -317,14 +338,18 @@ def compute_support_vals(self, r: torch.Tensor, phi: torch.Tensor, r_cutoff: flo
...
+        harmonic = torch.where(n % 2 == 1, torch.sin(torch.ceil(n_f / 2) * math.pi * x / width), torch.cos(torch.ceil(n_f / 2) * math.pi * x / width))
+        harmonic *= torch.where(m % 2 == 1, torch.sin(torch.ceil(m_f / 2) * math.pi * y / width), torch.cos(torch.ceil(m_f / 2) * math.pi * y / width))
 
         # computes the envelope
         vals = self.hann_window(r, width=width) * harmonic
 
-        # L2 normalization (skip during the initial norm computation in __init__)
-        if hasattr(self, "_l2_norms"):
+        # L2 normalization (skip during the initial norm computation in __init__, when _l2_norms is still None)
+        if self._l2_norms is not None:
```
bonevbs, line 355: isn't this more fragile then the 'hasattr' pattern?

bonevbs (commented):
I left a few comments, please have a look

tests/test_convolution.py
azrael417: I disagree. this recomputes something which is used over and over again, this should speed up the process by vectorization.

torch_harmonics/disco/convolution.py
azrael417: this is bypassing the basis norm, because we do not need to run through the whole loop

torch_harmonics/distributed/primitives.py
azrael417: this is a fix when using 1xN or Nx1 decompose, otherwise the routine would crash. I had that fix in the distributed attention branch and ported it over, otherwise the distributed tests for that would fail

tests/test_convolution.py
bonevbs: why is that so complicated? Wouldn't it be sufficient to just check not use AMP in the parameterized class when we run a CPU test?

azrael417: fixed with custom autocast

torch_harmonics/filter_basis.py
azrael417: this is only used for the spicy pass, no?

torch_harmonics/filter_basis.py
azrael417: no, there are cases where it was set but None.

torch_harmonics/filter_basis.py
azrael417: we can remove that, not necessary

torch_harmonics/filter_basis.py
```diff
@@ -184,11 +202,12 @@ def _compute_support_vals_isotropic(self, r: torch.Tensor, phi: torch.Tensor, r_
         nr = self.kernel_shape[0]
         dr = 2 * r_cutoff / (nr + 1)
 
-        # compute the support
+        # compute the support (cast to r.dtype so fp64 callers stay in fp64)
+        ikernel_f = ikernel.to(r.dtype)
```
bonevbs: do we need to rename this? why not just keep calling this ikernel?

torch_harmonics/filter_basis.py
```diff
@@ -489,23 +520,27 @@ def kernel_size(self) -> int:
...
         alphas = self._alphas
 
-        j_next = torch.tensor([float(scipy_jn(int(mi) + 1, float(a))) for mi, a in zip(ms, alphas)], dtype=torch.float64)
+        j_next = torch.tensor([float(_scipy_jn(int(mi) + 1, float(a))) for mi, a in zip(ms, alphas)], dtype=torch.float64)
         radial_norm = (j_next.abs() / math.sqrt(2)).clamp(min=1e-12)
         sqrt_2pi = math.sqrt(2 * math.pi)
         sqrt_pi = math.sqrt(math.pi)
         angular_sqrt = torch.where(ms == 0, torch.full_like(ms, sqrt_2pi), torch.full_like(ms, sqrt_pi))
-        return (radial_norm * angular_sqrt).clamp(min=1e-12)
+        return (r_cutoff * radial_norm * angular_sqrt).clamp(min=1e-12)
```
bonevbs, line 545: why do we multiply with the cutoff?

torch_harmonics/filter_basis.py
azrael417: fixed.

torch_harmonics/filter_basis.py
azrael417: so that it is similar to the behavior of the other BF

bonevbs: I am not sure this is correct - does this give the same result as before?

azrael417: it should, because it multiplies with one

bonevbs (approved):
LGTM

PR #175, opened 2026-04-24
applying midpoint rule for radial integral

azrael417:
This changes the radial integral in the basis function normalization computation to the midpoint rule with correct endpoint handling.


bonevbs (approved):
LGTM

PR #161, opened 2026-03-30
Distributed Neighborhood Attention S2

azrael417:
This MR adds distributed Neighborhood Attention S2 support and fixes some issues in the existing attention kernel.

- existing serial attention kernel preallocated an output tensor which was too large when using attention based downsampling. This is fixed
- this kernel computes numerically the same for self-attention and cross-attention with downsampling.
- this MR adds distributed neighborhood attention along with some new tests for the feature. This kernel does not support up- or downsampling yet

PR #176, opened 2026-04-27
Tkurth/attention upsampling

azrael417:
This MR fully implements cross attention for the serial codepath. For distributed attention, the upsampling attention path is still missing and will follow later. Feel free to review.

Should be merged after https://github.com/NVIDIA/torch-harmonics/pull/161


torch_harmonics/attention/attention.py
```diff
@@ -270,25 +270,35 @@ def __init__(
...
         self.optimized_kernel = optimized_kernel and optimized_kernels_is_available()
 
-        # heuristic to compute theta cutoff based on the bandlimit of the input field and overlaps of the basis functions
+        # heuristic to compute theta cutoff based on the bandlimit of the input field
+        # and overlaps of the basis functions. For upsample we follow DISCO's transpose
+        # convention and use the coarser (input) grid spacing.
         if theta_cutoff is None:
-            self.theta_cutoff = torch.pi / float(self.nlat_out - 1)
+            if self.upsample:
+                self.theta_cutoff = torch.pi / float(self.nlat_in - 1)
```
bonevbs, line 292: I think we can import the theta_cutoff routine tat we ue for convolution?

bonevbs (commented):
Looks mostly good to me, just one comment

torch_harmonics/attention/attention.py
azrael417: that is also not optimal. The issue is that the current routine uses the coarser lattice or the input lattice, ideally we should use the finer lattice for theta calculation. We can implement one central routine and use that everywhere.

azrael417: Let's do that in a separate MR though, because it touches a lot of routines not just this one

azrael417: tracking this here https://github.com/NVIDIA/torch-harmonics/issues/184

mauro-bis (approved):
The new upsample serial kernels (fwd+bwd) look good to me.

PR #177, opened 2026-04-28
adding pt2 compat tags and tests for the ring attention kernels

azrael417:
This MR adds pt2 compatible tags to the ring exchange kernels so that those can be lowered by inductor. It also adds tests for these kernels that verify inductor compatibility.


apaaris (approved):
Great, lgtm thanks

PR #178, opened 2026-04-29
initialization speedup part I

azrael417:
This MR adds vectorized routines for precompute_legpoly and precompute_dlegpoly with speedups up to 100x depending on size (5x on very small grids, 100x on larger grids). The old Legendre computation routines are moved into legendre_legacy.py and renamed accordingly. Tests are added to compute correctness.


torch_harmonics/legendre_legacy.py
bonevbs, line 1: I am not a fan of keeping legacy code alive in the library. Can we move to the new code?

bonevbs (changes requested):

torch_harmonics/legendre_legacy.py
azrael417: I kept is mainly for testing purposes, because it is a ground truth which works. But on the other hand, the old ground truth was also not tested against some other ground truth only against some special known cases. I can retire it if you like.

azrael417: gone.

bonevbs (approved):
LGTM

PR #179, opened 2026-04-30
initialization speedup part II

azrael417:
This MR uses vectorized routines in order to normalize the entries for the disco convolution tensor instead of using the double loop over kernels and latitude dim. I did not measure the exact speedup but the related tests run much faster. The correctness of the new tensor is validated by comparing to the old tensor element wise with 1e-9 precision.


tests/test_convolution.py
```diff
@@ -41,7 +41,11 @@
 
 from torch_harmonics.quadrature import precompute_latitudes, precompute_longitudes
 from torch_harmonics.disco import cuda_kernels_is_available, optimized_kernels_is_available
-from torch_harmonics.disco.convolution import _precompute_convolution_tensor_s2
+from torch_harmonics.disco.convolution import (
+    _precompute_convolution_tensor_s2,
+    _normalize_convolution_tensor_s2,
+    _normalize_convolution_tensor_s2_legacy,
```
bonevbs: could you please also remove legacy here?

torch_harmonics/disco/convolution.py
```diff
@@ -61,7 +61,134 @@ def _normalize_convolution_tensor_s2(
...
+
+    # scale[ik, ilat] = sum |psi_vals - bias[ik, ilat]| * q over the neighborhood
+    bias_per_nz = bias.view(-1)[gid]
+    scale_flat = torch.zeros(n_groups, dtype=psi_vals.dtype, device=psi_vals.device)
+    scale_flat.scatter_add_(0, gid, (psi_vals - bias_per_nz).abs() * q)
+    scale = scale_flat.view(kernel_size, nlat_out)
+
+    # per-mode (b, s) selection per nonzero, then renormalize in a single elementwise pass
+    if basis_norm_mode in ("nodal", "modal"):
+        b_per_nz = bias.view(-1)[gid]
```
bonevbs: can you write the variable names out? I think scale_per_nz is more understable

bonevbs (commented):
Mainly delete the legacy code and make the variable naming more understandable.

tests/test_convolution.py
azrael417: done

torch_harmonics/disco/convolution.py
azrael417: I changed this but review carefully, this was not entirely trivial since similar names are used in similar places

torch_harmonics/disco/convolution.py
```diff
@@ -61,55 +61,11 @@ def _normalize_convolution_tensor_s2(
...
-    Returns
-    -------
-    torch.Tensor
-        Normalized convolution tensor values.
-
-    Raises
-    ------
-    ValueError
-        If basis_norm_mode is not one of the supported modes.
+    Each nonzero is assigned a flat group id ``gid = ikernel * nlat_out + ilat_out``
```
bonevbs: ewhy was the docstring shortened so drastically? In addition, I do not understand this new docstring. The old one was more readable.

tests/test_convolution.py
```diff
@@ -41,7 +41,10 @@
 
 from torch_harmonics.quadrature import precompute_latitudes, precompute_longitudes
 from torch_harmonics.disco import cuda_kernels_is_available, optimized_kernels_is_available
-from torch_harmonics.disco.convolution import _precompute_convolution_tensor_s2
+from torch_harmonics.disco.convolution import (
+    _precompute_convolution_tensor_s2,
+    _normalize_convolution_tensor_s2,
+)
```
bonevbs, line 47: could you please also remove legacy here?

azrael417: where is legacy here?

torch_harmonics/disco/convolution.py
azrael417: fixed :)

torch_harmonics/filter_basis.py
```diff
@@ -356,10 +356,13 @@ def compute_support_vals(self, r: torch.Tensor, phi: torch.Tensor, r_cutoff: flo
...
 
-        # L2 normalization (skip during the initial norm computation in __init__, when _l2_norms is still None)
+        # L2 normalization (skip during the initial norm computation in __init__, when _l2_norms is still None).
+        # The 1/r_cutoff factor makes ||psi||_{L^2(disk of radius r_cutoff)} = 1 for any r_cutoff: _l2_norms
+        # are computed on the unit disk, and substituting u = r/r_cutoff in the L^2 integral gives a r_cutoff^2
+        # Jacobian, which the 1/r_cutoff prefactor (along with the unit-disk norm) cancels.
         if self._l2_norms is not None:
             norms = self._l2_norms.to(device=vals.device, dtype=vals.dtype)
-            vals = vals / norms[iidx[:, 0]].clamp(min=1e-12)
+            vals = vals / (norms[iidx[:, 0]].clamp(min=1e-12) * r_cutoff)
```
bonevbs: is this consistent with what we do in Fourier-Bessel and in Piecewise linear?

azrael417: we got tests which test that, Fourier Bessel is somehow correct by default, the others need that factor.

azrael417: ah, PL is not tested. let me check if we can expect that. PL is tricky since it has uneven support.

azrael417: ah, PL is a partition of unity and that is tested separately. applying L2 normalization would destroy that property. However, all other BF obey that

azrael417: added a comment about that.

bonevbs (dismissed):
LGTM - I have tested both the init and the filter basis normalization

torch_harmonics/filter_basis.py
bonevbs: I have moved to the unit disk convention. The normalization which takes the radius into account happens at the psi tensor normalization level. This is also beneficial for future implementations of wavelets and lifting schemes

apaaris (dismissed):
Looks good, thanks lgtm

bonevbs (approved):
LGtM

PR #180, opened 2026-05-06
Conv adapter for different numbers of input and output channels

azrael417:
This MR adds a conv adapter for the residual pass in SFNO when input and output channels are different.


bonevbs (approved):
LGTM

PR #183, opened 2026-05-06
making docstring with escapes raw

azrael417:
this MR should fix the torch-harmonics/torch_harmonics/filter_basis.py:449: SyntaxWarning: invalid escape sequence '\p' warning


bonevbs (approved):
LGTM

PR #186, opened 2026-05-11
adding specialization from disco bwd pass to fwd pass for pscale = 1,2,3

azrael417:
This MR adds the same template specialization to disco FWD than what we have for bwd. This gives a moderate speedup of 10-15% depending on GPU arch.


bonevbs (approved):
LGTM

PR #187, opened 2026-05-12
cleaning up attention tests

azrael417:
This MR removes unused function arguments to some optimized kernels and also adds additional tests for the attention gradient contract


bonevbs (approved):
LGTM

PR #188, opened 2026-05-12
DISCO tests for gradient contract

azrael417:
This MR adds tests to check whether disco returns None grads when no grads are required.


bonevbs (approved):
LGTM

PR #189, opened 2026-05-12
gradient computation fix

azrael417:
this MR skips the computation of not required gradients in distributed attention. it also adds tests to test the correctness.

### DIFF
--- a/torch_harmonics/distributed/distributed_attention.py
+++ b/torch_harmonics/distributed/distributed_attention.py
@@ -195,13 +195,13 @@ class _RingNeighborhoodAttentionFn(torch.autograd.Function):
         # of {kw, vw, qw} that doesn't need a gradient, and return None in those slots.
         # This is what lets torch.compile / AOTAutograd prune dead subgraphs (including
         # the NCCL allreduces) from the compiled backward.
-        kw_needs = ctx.needs_input_grad[0]
-        vw_needs = ctx.needs_input_grad[1]
-        qw_needs = ctx.needs_input_grad[2]
+        kw_needs_grad = ctx.needs_input_grad[0]
+        vw_needs_grad = ctx.needs_input_grad[1]
+        qw_needs_grad = ctx.needs_input_grad[2]
 
         # Defensive: if somehow none of (kw, vw, qw) need grad (e.g., user wired
         # requires_grad onto one of the index buffers), there's nothing to compute.
-        if not (kw_needs or vw_needs or qw_needs):
+        if not (kw_needs_grad or vw_needs_grad or qw_needs_grad):
             return (None,) * 18
 
         B, C_k, H_halo, _ = kw.shape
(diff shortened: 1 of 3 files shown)


torch_harmonics/distributed/primitives.py
```diff
@@ -206,18 +206,19 @@ def _reduce(input_, use_fp32=True, group=None):
...
     if use_fp32:
         dtype = input_.dtype
-        inputf_ = input_.float()
-        inputf_ = inputf_.contiguous()
+        inputf_ = input_.float().contiguous().clone()
         dist.all_reduce(inputf_, group=group)
         input_ = inputf_.to(dtype)
     else:
-        input_ = input_.contiguous()
+        input_ = input_.contiguous().clone()
```
bonevbs, line 219: I see this was already the case with fp32 before. What's the performance hit of the clone here and ewhy is this necessary here?

torch_harmonics/distributed/distributed_attention.py
```diff
@@ -191,6 +191,19 @@ def backward(ctx, dy, _dalpha_sum, _dqdotk_max):
         az_rank          = ctx.az_rank
         az_size          = ctx.az_size
 
+        # Autograd contract: skip per-branch work (kernel calls, allreduces) for any
+        # of {kw, vw, qw} that doesn't need a gradient, and return None in those slots.
+        # This is what lets torch.compile / AOTAutograd prune dead subgraphs (including
+        # the NCCL allreduces) from the compiled backward.
+        kw_needs = ctx.needs_input_grad[0]
```
bonevbs: can we rename those to `qw_needs_grad` etc? I know this is longer but i prefer the verbose names.

torch_harmonics/distributed/distributed_attention.py
```diff
@@ -239,81 +255,108 @@ def backward(ctx, dy, _dalpha_sum, _dqdotk_max):
...
+        if kw_needs or vw_needs:
+            integral_norm = integral_buf * alpha_sum_inv                  # [B, H, W]
 
         # dqy[b,h,w,c] = inv_sq*(alpha_sum*alpha_kvw - integral*alpha_k)
-        dqy_cl = alpha_sum_inv_sq.unsqueeze(-1) * (
-            fwd_alpha_sum.unsqueeze(-1) * alpha_kvw_buf
-            - integral_buf.unsqueeze(-1) * alpha_k_buf
-        )                                                                 # [B, H, W, C_k]
-        dqy = dqy_cl.permute(0, 3, 1, 2).contiguous()                  # [B, C_k, H, W]
+        if qw_needs:
```
bonevbs: does this make a big difference in terms of performance? I am wondering if it wouldn't be fine to simply compute all gradients by default. On the other hand omitting this might save quite a bit of memory?

bonevbs (commented):
Some comments here and there. Please have a look

torch_harmonics/distributed/primitives.py
azrael417: we only had the clone in the non fp32 case, the clone in fp32 is super important if the data is in fp32 already because otherwise it would be modified in-place which is wrong. Performance is worse in some cases but this is a necessary fix.

torch_harmonics/distributed/distributed_attention.py
azrael417: afaiu, this is the modern style. It saves performance and memory. And it is required now to check which gradients to return and None has to be returned if the input does not need a gradient, otherwise torch.compile correctness cannot be guaranteed.

torch_harmonics/distributed/distributed_attention.py
azrael417: I did not change the name, can we leave them for the time being, changing those can introduce new issues I want to avoid atm.

bonevbs: With new issues you mean bugs?

bonevbs (approved):
LGTM

PR #192, opened 2026-05-13
Tkurth/disco cpu improvements

azrael417:
This MR improves the OpenMP kernels for DISCO with about 30x speedup on average. They more closely mirror the CUDA kernels now, which makes then also easier to maintain. Also, it improves execution time of tests and CPU inference workloads.


bonevbs (approved):
LGTM, feel free to merge when the new kernel passes all unit tests

PR #193, opened 2026-05-15
removing in-place modifications in inverse fft

azrael417:
This MR fixes an in-place modification issue which throws MKL-FFT off in the backward pass.

### DIFF
--- a/torch_harmonics/fft.py
+++ b/torch_harmonics/fft.py
@@ -74,15 +74,19 @@ def irfft(x: torch.Tensor, n: Optional[int] = None, dim: int = -1, **kwargs) -> 
     if n is None:
         n = 2 * (x.size(dim) - 1)
 
-    # ensure that imaginary part of 0 and nyquist components are zero
-    # this is important because not all backend algorithms provided through the
-    # irfft interface ensure that
-    if x.requires_grad:
-        # this ensures that we create a contiguous new tensor we can overwrite:
-        x = x.clone()
-    x[..., 0].imag = 0.0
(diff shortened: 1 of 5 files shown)


torch_harmonics/fft.py
```diff
@@ -74,16 +76,20 @@ def irfft(x: torch.Tensor, n: Optional[int] = None, dim: int = -1, **kwargs) ->
...
+    # on CPU+MKL (DftiCommitDescriptor rejects the resulting stride pattern).
+    imag_scale = torch.ones(x.size(dim), dtype=x.real.dtype, device=x.device)
+    imag_scale[0] = 0.0
     if (n % 2 == 0) and (n // 2 < x.size(dim)):
-        x[..., n // 2].imag = 0.0
+        imag_scale[n // 2] = 0.0
+    shape = [1] * x.ndim
+    shape[dim] = -1
+    imag_scale = imag_scale.reshape(*shape)
+    x = torch.complex(x.real, x.imag * imag_scale)
```
bonevbs, line 91: Does this imply extra memory in the backwards pass? Just wondering if there is a simpler, MKL-compatible way to do this here.

bonevbs (commented):
One high-Level comment

torch_harmonics/fft.py
azrael417: it does increase memory requirements but in place modifications we should generally avoid because it can cause very subtle and nasty bugs. Therefore I think we should merge this. The fact that MKL triggers it is actually good because otherwise we wouldn't have found it.

bonevbs (approved):
LGTM

PR #194, opened 2026-05-15
wiring up equidistant

azrael417:
this MR wires up equidistant for the quadrature tests.

### DIFF
--- a/torch_harmonics/quadrature.py
+++ b/torch_harmonics/quadrature.py
@@ -126,8 +126,8 @@ def trapezoidal_weights(n: int, a: Optional[float]=-1.0, b: Optional[float]=1.0,
         Tensor of quadrature weights
     """
 
-    xlg = torch.as_tensor(np.linspace(a, b, n, endpoint=periodic))
-    wlg = (b - a) / (n - periodic * 1) * torch.ones(n, requires_grad=False)
+    xlg = torch.as_tensor(np.linspace(a, b, n, endpoint=not periodic))
+    wlg = (b - a) / (n - 1 + periodic * 1) * torch.ones(n, requires_grad=False)
 
     if not periodic:
         wlg[0] *= 0.5
--- a/torch_harmonics/quadrature.py
+++ b/torch_harmonics/quadrature.py
@@ -389,6 +389,11 @@ class QuadratureS2(torch.nn.Module):
             dlambda = 2 * torch.pi / img_shape[1]
             quad_weight = dlambda * weights.unsqueeze(1)
             quad_weight = quad_weight.tile(1, img_shape[1])
+        elif self.grid == "equidistant":
+            _, weights = trapezoidal_weights(img_shape[0], -1, 1)
+            dlambda = 2 * torch.pi / img_shape[1]
+            quad_weight = dlambda * weights.unsqueeze(1)
+            quad_weight = quad_weight.tile(1, img_shape[1])
         else:
             raise (ValueError("Unknown quadrature mode"))
 
--- a/torch_harmonics/quadrature.py
+++ b/torch_harmonics/quadrature.py
@@ -288,54 +288,6 @@ def clenshaw_curtiss_weights(n: int, a: Optional[float]=-1.0, b: Optional[float]
     return tcc, wcc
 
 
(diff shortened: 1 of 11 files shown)


torch_harmonics/quadrature.py
```diff
@@ -126,8 +126,8 @@ def trapezoidal_weights(n: int, a: Optional[float]=-1.0, b: Optional[float]=1.0,
         Tensor of quadrature weights
     """
 
-    xlg = torch.as_tensor(np.linspace(a, b, n, endpoint=periodic))
-    wlg = (b - a) / (n - periodic * 1) * torch.ones(n, requires_grad=False)
+    xlg = torch.as_tensor(np.linspace(a, b, n, endpoint=not periodic))
+    wlg = (b - a) / (n - 1 + periodic * 1) * torch.ones(n, requires_grad=False)
```
bonevbs, line 130: is this correct? This seems only cerrect if you assume that the endpoints are included.

bonevbs (commented):
minor question

torch_harmonics/quadrature.py
azrael417: it is an open interval if the interval is periodic, so in that case the endpoints need to be dropped.

I am happy with removing the weights entirely though since we do not use them and quadrature errors are 4 orders of magnitude larger than other comparable weights.

bonevbs: Hmm is this for the longitude domain or for the latitude? If I understand correctly, this is equidistant in $\cos \theta$? If it is in the latitude domain endpoints should be included. This is effectively an alternative to Clenshaw-Curtiss quadrature but there is no upside to the best of my knowledge aside from the fact that everyone is familiar with trapezoidal rule.

I would still keep them for legacy reasons. Some very early version of torch-harmonics was using these for the projection onto the Legendre basis if I recall correctly.

azrael417: this is for both domains, and for lat you set periodic to false and for Lon periodic to true. and in the former the endpoint is included and in the latter it is not, which I correct I think. The error is much smaller with this setting than in the old case.

torch_harmonics/quadrature.py
```diff
@@ -389,6 +341,11 @@ def __init__(
             dlambda = 2 * torch.pi / img_shape[1]
             quad_weight = dlambda * weights.unsqueeze(1)
             quad_weight = quad_weight.tile(1, img_shape[1])
+        elif self.grid == "equidistant":
```
bonevbs: is this due to legacy compatibility with makani? I feel `equidistant` and `equiangular` might confuse people.

azrael417: fixed

torch_harmonics/quadrature.py
```diff
@@ -35,7 +35,7 @@
 import numpy as np
 import torch
 
-def _precompute_quadrature_weights(n: int, grid: Optional[str]="equidistant", a: Optional[float]=0.0, b: Optional[float]=1.0,
+def _precompute_quadrature_weights(n: int, grid: Optional[str]="equiangular-trapezoidal", a: Optional[float]=0.0, b: Optional[float]=1.0,
```
bonevbs: why is this by default equidistant/equianfular-trapezoidal?

torch_harmonics/quadrature.py
```diff
@@ -45,13 +45,13 @@ def _precompute_quadrature_weights(n: int, grid: Optional[str]="equidistant", a:
     n : int
         Number of grid points
     grid : str, optional
-        Grid type ("equidistant", "legendre-gauss", "lobatto", "equiangular"), by default "equidistant"
+        Grid type ("equiangular-trapezoidal", "legendre-gauss", "lobatto", "equiangular"), by default "equiangular-trapezoidal"
```
bonevbs: also here

bonevbs (commented):
i am puzzled why the default is/was `equiangular-trapezoidal/equidistant` in `quadrature.py`.

torch_harmonics/quadrature.py
azrael417: fixed

torch_harmonics/quadrature.py
azrael417: fixed

bonevbs (approved):
LGTM

PR #195, opened 2026-05-18
0.9.2: attention cpu improvements

azrael417:
This MR improves the performance of all the CPU-based attention kernels.


bonevbs (approved):
LGTM

PR #196, opened 2026-05-18
0.9.2: Fused Disco kernel support, reducing memory footprint by a factor of K 

azrael417:
This MR adds the support for a fused disco kernel which helps reduce memory pressure for single gpu training


tests/test_convolution.py
```diff
@@ -457,6 +466,7 @@ def test_sparse_against_dense(
         grid_out,
         dtype,
         transpose,
+        fused,
```
bonevbs, line 469: all of the tests use the non fused kernekl, intentional?

tests/test_convolution.py
```diff
@@ -695,6 +719,7 @@ def test_optimized_against_torch(
             bias=False,
             theta_cutoff=theta_cutoff,
             optimized_kernel=True,
+            **fused_kwarg,
```
bonevbs, line 722: do we need to support both fused and unfused? My preference would be to fuse it in principle

tests/test_convolution.py
azrael417: all the old tests use fused=false, and then I added new ones with fused=True

tests/test_convolution.py
azrael417: I think it is good to keep the unfused around, as it is slightly faster

bonevbs (approved):
LGTM

PR #197, opened 2026-05-18
0.9.2: Fixing Reduce scatter and wiring it into distributed disco

azrael417:
This MR fixes reduce-scatter for uneven shapes across ranks by employing padding before the call and removing the padding. This should cut down comm time in the allreduce phase of the disorder kernel by a factor of number of ranks in H. 

I also added a test which verifies the new reduce-scatter. 

Furthermore, some torch sparse tensor constructions do have disabled range checks to suppress warnings.


torch_harmonics/filter_basis.py
```diff
@@ -124,7 +124,12 @@ def compute_l2_norms(self, r_cutoff: float = 1.0, nr: int = 50, nphi: int = 200)
         r, phi = torch.meshgrid(r, phi, indexing="ij")
 
         iidx, vals = self.compute_support_vals(r, phi, r_cutoff=r_cutoff)
-        psi = torch.sparse_coo_tensor(iidx.t(), vals, size=(self.kernel_size, nr, nphi)).to_dense()
+        # Indices come from compute_support_vals and are valid by construction.
+        # Skip the sparse-invariant pass and silence the implicit-default
+        # UserWarning PyTorch emits when neither enable=True nor enable=False
+        # is chosen explicitly.
+        with torch.sparse.check_sparse_tensor_invariants(enable=False):
```
bonevbs, line 131: aren't these disabled by default?

azrael417: this gets rid off a warning which would otherwise show up related to those.

bonevbs (approved):
LGTM - one question but feel free to merge - i guess it makes sense

PR #200, opened 2026-05-19
Bbonev/contributing guide

bonevbs:
Adding a Contributing guide. Also bumping the version number to 0.9.0. Switching `python` to `python3` on some internal script to be compatible with the modern Linux convention where `python` may not exist.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 Thank you for your interest in contributing. torch-harmonics implements differentiable
 signal processing on the sphere (SHT, DISCO convolutions, spherical attention, and
-distributed variants). We are gratefule for contributions that improve correctness, performance,
+distributed variants). We are grateful for contributions that improve correctness, performance,
 documentation, or test coverage.
 
 ## Table of contents
@@ -108,7 +108,7 @@ export TORCH_CUDA_ARCH_LIST="8.0 8.6 9.0"   # set to GPUs you target; reduces co
 pip install -e . --no-build-isolation
 ```
 
-Custom CUDA extensions require compute capability **≥ 7.0**.
+Custom CUDA extensions require compute capability **≥ 8.0**.
 
 ### Build environment variables
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,327 @@
+# Contributing to torch-harmonics
+
+Thank you for your interest in contributing. torch-harmonics implements differentiable
+signal processing on the sphere (SHT, DISCO convolutions, spherical attention, and
+distributed variants). We are gratefule for contributions that improve correctness, performance,
+documentation, or test coverage.
+
+## Table of contents
+
+- [Getting in touch](#getting-in-touch)
+- [Development setup](#development-setup)
+- [Building C++/CUDA extensions](#building-ccuda-extensions)
+- [Running tests](#running-tests)
+- [Code style and pre-commit](#code-style-and-pre-commit)
+- [Project structure](#project-structure)
+- [Guidelines by area](#guidelines-by-area)
+- [Pull requests](#pull-requests)
+- [Release and packaging](#release-and-packaging)
+
+## Getting in touch
+
+We're happy to discuss ideas before you spend time on a large change.
+
+- Open a [GitHub issue](https://github.com/NVIDIA/torch-harmonics/issues) for bugs,
+  feature proposals, or design questions. A similar effort may already be in progress.
+- For larger changes (new APIs, breaking behavior, distributed design), an issue first
+  helps align on approach. Small fixes and test improvements can often go straight to a PR.
+- Please be respectful and constructive in issues and reviews.
+
+## Development setup
+
+**Requirements**
+
+- Python 3.9+
+- PyTorch 2.6+ (install before building; extensions compile against your local `torch`)
+- NumPy 1.22.4+
+- A C++17 compiler; CUDA toolkit optional but recommended for GPU kernel work
+
+**Editable install**
+
+```bash
+git clone https://github.com/NVIDIA/torch-harmonics.git
+cd torch-harmonics
+
+# Install PyTorch first (CPU example; use the CUDA wheel that matches your system)
+python3 -m pip install torch --extra-index-url https://download.pytorch.org/whl/cpu
+
+# Editable install with dev dependencies
+python3 -m pip install -e ".[dev]" --no-build-isolation
+```
+
+Optional extras:
+
+```bash
+pip install -e ".[dev,filter_basis]"   # scipy for filter-basis tests
+```
+
+Install [pre-commit](https://pre-commit.com/) hooks (see [Code style and pre-commit](#code-style-and-pre-commit)):
+
+```bash
+pre-commit install
+```
+
+## Building C++/CUDA extensions
+
+torch-harmonics ships several compiled extensions:
+
+| Module | Purpose |
+|--------|---------|
+| `torch_harmonics.attention._C` | Optimized neighborhood attention (CPU/CUDA) |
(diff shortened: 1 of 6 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,339 @@
...
+  feature proposals, or design questions. A similar effort may already be in progress.
+- For larger changes (new APIs, breaking behavior, distributed design), an issue first
+  helps align on approach. Small fixes and test improvements can often go straight to a PR.
+- Please be respectful and constructive in issues and reviews.
+
+## Development setup
+
+**Requirements**
+
+- Python 3.9+
```
azrael417, line 34: this should be 3.10

CONTRIBUTING.md
```diff
@@ -0,0 +1,339 @@
...
+- PyTorch 2.6+ (install before building; extensions compile against your local `torch`)
+- NumPy 1.22.4+
```
azrael417, line 36: I think we nailed this to 2.3+. please check

pyproject.toml
```diff
@@ -55,7 +55,8 @@ dev = [
     "coverage>=6.5.0",
     "parameterized",
     "scipy>=1.9.0",
-    "pre-commit>=3.0.0",
+    "pre-commit>=4.0.0",
+    "virtualenv>=20.23.0",
```
azrael417: why do we enforce the pre-commit to run in a venv?

CONTRIBUTING.md
```diff
@@ -0,0 +1,339 @@
...
+- For larger changes (new APIs, breaking behavior, distributed design), an issue first
+  helps align on approach. Small fixes and test improvements can often go straight to a PR.
+- Please be respectful and constructive in issues and reviews.
+
+## Development setup
+
+**Requirements**
+
+- Python 3.9+
+- PyTorch 2.6+ (install before building; extensions compile against your local `torch`)
```
azrael417, line 35: in setup.py we set >= 2.8.0 and pyproject we set 2.6.0. We should maybe fix this so the same number.

CONTRIBUTING.md
```diff
@@ -0,0 +1,339 @@
...
+### Building wheels locally
+
+```bash
+python3 -m pip install build
+python3 -m build --wheel --no-isolation
+```
+
+Wheels follow the PyTorch ecosystem pattern
+`torch_harmonics-{version}+{cuda}-{python}-{abi}-{platform}.whl`, e.g.
+`torch_harmonics-0.9.1+cu124-cp310-cp310-linux_x86_64.whl`.
```
azrael417: cu124 is not supported, better use cu126

CONTRIBUTING.md
```diff
@@ -0,0 +1,339 @@
...
+
+If CUDA is not detected automatically (containers, headless nodes):
+
+```bash
+export FORCE_CUDA_EXTENSION=1
+export TORCH_CUDA_ARCH_LIST="8.0 8.6 9.0"   # set to GPUs you target; reduces compile time
+pip install -e . --no-build-isolation
+```
+
+Custom CUDA extensions require compute capability **≥ 7.0**.
```
azrael417: maybe we should bump this to 8.0, since that is also what the newer NGC containers have and our Dockerfile.

CONTRIBUTING.md
```diff
@@ -0,0 +1,339 @@
+# Contributing to torch-harmonics
+
+Thank you for your interest in contributing. torch-harmonics implements differentiable
+signal processing on the sphere (SHT, DISCO convolutions, spherical attention, and
+distributed variants). We are gratefule for contributions that improve correctness, performance,
```
azrael417: grateful

CONTRIBUTING.md
```diff
@@ -0,0 +1,339 @@
...
+│   └── kernels_torch/                                               # Torch reference implementations
+└── distributed/                                                     # Multi-GPU primitives and layers
+
+tests/              # Pytest suite (shared helpers in testutils.py)
+examples/           # Training and usage examples (not run in default CI)
+notebooks/          # Exploratory notebooks (not run in CI)
+```
+
+**Dual implementation rule:** If optimized kernels are implemented, torch reference implementations must be
+provided to ensure readability and their ouputs must agree within test tolerances. If you change kernel semantics,
```
azrael417: outputs

CONTRIBUTING.md
```diff
@@ -0,0 +1,339 @@
...
+update both paths (or the shared sparsity / indexing logic) and add or extend tests.
```
azrael417: we should be a bit more generous here. relaxing test tolerances can be OK if the algorithm is different but we should require that this should be highlighted and specifically spelled out in the changelog.

CONTRIBUTING.md
```diff
@@ -0,0 +1,339 @@
...
+  - **Upsample:** `nlon_out % nlon_in == 0` (serial `NeighborhoodAttentionS2`; distributed
+    upsample not implemented yet)
+- After editing attention C++ or CUDA sources, rebuild and run `tests/test_attention.py`
+  (including upsample cases).
+
+### Distributed
+
+- Use `TORCH_HARMONICS_DISTRIBUTED_DEBUG` for extra shape checks (see distributed module
+  docs).
+- Ring-attention and halo-exchange paths have dedicated op tests with
```
azrael417: in general, for all custom kernels we require implementation of pt2 compatibility tests using opcheck.

CONTRIBUTING.md
```diff
@@ -0,0 +1,339 @@
...
+
+1. **Branch** from `main` using `username/feature-name` (e.g. `jdoe/fourier-bessel-basis`).
+2. **Keep PRs focused.** One logical change per PR is easier to review.
+3. **Add tests** for bug fixes and new behavior.
+4. **Update `Changelog.md`** for user-visible changes, especially breaking ones.
+5. **Describe the PR clearly:** what problem it solves, how you tested it (commands,
+   CPU/GPU), and any API or numerical behavior changes.
+6. **Ensure CI passes:** style (pre-commit) and tests workflows.
+
+Linking to an open issue when one exists is helpful but not always required.
```
azrael417: I would drop the "always "

CONTRIBUTING.md
```diff
@@ -0,0 +1,339 @@
...
+
+Reviewers may ask for reference-kernel parity checks, justification for tolerance changes,
+or benchmarks for performance-critical kernel changes.
```
azrael417: we should implement the benchmarking logic to simplify this. We can add: Standardized benchmark is in preparation.

azrael417 (changes requested):
LGTM, but I have some suggestions

CONTRIBUTING.md
bonevbs: in the prproject.toml, we say torch-harmonics requires 3.9+.shouldn't I keep this consistent?

CONTRIBUTING.md
bonevbs: in the pyproject.toml it is 2.6+

azrael417: for numpy?

CONTRIBUTING.md
bonevbs: I changed setupy.py

pyproject.toml
bonevbs: do we enforce this? I removed venv from the dev requirements

CONTRIBUTING.md
bonevbs: I addressed thisin an extra section

CONTRIBUTING.md
azrael417: I checked, we support 3.9. We had a 3.10 feature in the code but I think ruff replaced this in his pass. So all good.

azrael417 (approved):
LGTM, thanks for setting that up.

PR #201, opened 2026-05-20
0.9.2: Ring exchange distributed DISCO kernel support and partial fusing for vastly reduced memory footprint

azrael417:
Distributed DISCO conv: ring-exchange backend

  Adds a ring-exchange algorithm for DistributedDiscreteContinuousConvS2, selected via a new method= flag (TE-style). Default "a2a" keeps current behavior byte-identical; "ring" and
  "ring", fused=True are new opt-ins.

  Why: the fused ring path drops peak_bwd substantially in the channel-heavy regime (≈35% on S3-style configs under bf16 autocast). Weight stays replicated; the K-expanded intermediate
   never crosses a collective.

  New surface area:
  - Two CUDA kernels (disco_cuda_{fwd,bwd}_ring.cu).
  - New torch_harmonics/distributed/kernels/ subpackage holding the orchestration.
  - Polar reduce_scatter wiring (used by both A2A and ring).
  - Opt-in pre-allocated P2P recv pool via TORCH_HARMONICS_P2P_BUFFER=1 (extra ~25% peak_bwd cut on non-fused ring; default off).
  
  Tests: tests/test_distributed_convolution.py extended to cover all (method, fused, p2p_buffer) combinations against the serial reference; 38 cases pass on the 2×2 harness.

Old method (="a2a") is unchanged; ring path is gated behind explicit opt-in.

This MR also adds dtype faithfulness to some custom ops, which is required for a follow up autogrid-eligibility pass. This will be done in a later MR but the faithfulness fixes are applied here because they are currently training progress.

Depends on: https://github.com/NVIDIA/torch-harmonics/pull/196 and https://github.com/NVIDIA/torch-harmonics/pull/197


torch_harmonics/distributed/primitives.py
```diff
@@ -644,6 +644,10 @@ def reduce_from_scatter_to_azimuth_region(input_, dim_):
     return _ReduceFromScatterToAzimuthRegion.apply(input_, dim_)
 
 
+@torch.compiler.disable()
```
bonevbs: this is duplicated now?

torch_harmonics/distributed/distributed_convolution.py
```diff
@@ -29,6 +29,7 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #
 
+import os
```
bonevbs, line 32: do we need os here? This is weird

azrael417: yes, because we have new environment variables one can set.

tests/test_distributed_convolution.py
```diff
@@ -29,14 +29,17 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 #
 
+import os
```
bonevbs, line 32: is os needed here?

azrael417: yes, since we have new env variables which change the behavior of some disco kernels, this setting allows us to run the tests in a mock env

torch_harmonics/distributed/primitives.py
azrael417: fixed

torch_harmonics/attention/kernels_torch/attention_torch.py
```diff
@@ -472,6 +476,17 @@ def _neighborhood_s2_attention_bwd_torch(ctx, grad_output):
     B, _, H, W = grad_output.shape
     grad_output = grad_output.reshape(B * nh, -1, H, W)
 
+    # Promote to fp32 internally for softmax numerics; cast each grad back to
+    # its corresponding input's dtype on return. Mirrors the optimized bwd
+    # pattern in attention_optimized.py.
+    kw_dtype = kw.dtype
+    vw_dtype = vw.dtype
+    qw_dtype = qw.dtype
+    kw = kw.to(torch.float32)
```
bonevbs, line 485: what if we get float64 input?

azrael417: we do not support FP64 attention yet. we can add this as an extra feature later. I will open a GitHub issue

azrael417: https://github.com/NVIDIA/torch-harmonics/issues/204

bonevbs (approved):
LGTM

Issue #202, opened 2026-05-22
Ensuring autocast eligibility for all custom ops


azrael417:
Currently, our custom ops are not autocast eligible. This can lead to problems where higher precision is used than necessary if the very first layer in a network is a TH custom op. We need to mark all custom ops as autograd eligible to circumvent this. This is a bigger effort, tracking it here.

azrael417:
This is now working.

PR #205, opened 2026-05-27
Tkurth/autocast correctness

azrael417:
This MR adds more thorough autocast correctness to the main branch. Attention kernels do not respect that yet but DISCO does.

also, this MR removes all references to torch.jit entirely to remove some deprecation warnings.

This would address https://github.com/NVIDIA/torch-harmonics/issues/202


tests/test_convolution.py
```diff
@@ -435,11 +435,11 @@ def test_convolution_tensor_integrity(self, in_shape, out_shape, kernel_shape, b
             [8, 4, 2, (12, 24), (24, 48), (2, 2), "harmonic", "mean", "equiangular", "equiangular", torch.float16, True, False, 2e-2, 1e-2],
             # bf16 tests (AMP)
             # regular convolution
-            [8, 4, 2, (16, 32), (16, 32), (3), "piecewise linear", "mean", "equiangular", "equiangular", torch.bfloat16, False, False, 5e-2, 5e-2],
-            [8, 4, 2, (24, 48), (12, 24), (2, 2), "harmonic", "mean", "equiangular", "equiangular", torch.bfloat16, False, False, 5e-2, 5e-2],
+            [8, 4, 2, (16, 32), (16, 32), (3), "piecewise linear", "mean", "equiangular", "equiangular", torch.bfloat16, False, False, 2e-1, 5e-2],
```
bonevbs, line 438: is this normal that these tolerance are so high?

azrael417: yes. bfloat is very imprecise in this regard. it is super sensitive to oder of summations so the tolerances need to be relaxed a lot. this is similar to what TE does as well.

bonevbs (approved):
LGTM

PR #213, opened 2026-06-03
distributed sht improvements

### LINKED ISSUE #207: Distributed SHT fails when channel count is smaller than polar group size
## Summary

`torch_harmonics.distributed` appears to require the channel dimension to be at least as large as the polar process group size.

For scalar fields with `num_chans = 1`, distributed SHT can fail when `comm_size_polar > 1` with an error like:

`cannot split dim -3 of size 1 into K chunks`

This can arise in downstream code that applies distributed SHTs to individual 2-D scalar fields, where the input shape is adapted to `(channel, nlat, nlon)` and the effective channel count is one.

## Workaround

A local workaround is to zero-pad the channel axis up to `comm_size_polar` before calling the distributed transform, then slice the output back to the original channel count.

Since the SHT is linear, the padded zero channels remain zero and the original channels are unchanged.

## Request

Would you consider one of the following?

1. Handling `num_chans < comm_size_polar` internally with padding.
2. Documenting this channel-count constraint for distributed transforms.
3. Adding a clearer error message for this case.

azrael417:
this MR adds more flexible options to perform distributed SHT by sharing along all dims except spatial and padding if that is not sufficient. Tests were added to cover these cases.

fixes https://github.com/NVIDIA/torch-harmonics/issues/207


bonevbs (approved):
lgtm

PR #214, opened 2026-06-09
Fused algorithm for distributed DISCO

azrael417:
This MR adds a fused algorithm for distributed DISCO which is more memory efficient than the older ring algorithm and also 15x faster. I removed the ring algorithm and the API is not the same for serial and distributed DISCO.


bonevbs (dismissed):
LGTM - Please extensively test it in distributed mode before merginhg

bonevbs (approved):
LGTM

PR #215, opened 2026-06-11
Fix basis norm standard in distributed DISCO

bonevbs:
resolving discrepancy in the default parameters between distributed and sequential DISCO implementations


azrael417 (changes requested):
Please add that change to the changeling since it is backwards compatible breaking.

azrael417 (approved):
LGTM

PR #216, opened 2026-06-12
adding row-split caching

azrael417:
for performance reasons, the kernel re-computes the dense vs sparse splits with every pass. this is not necessary since this split is constant. This MR moves this into the constructor, improving kernel performance and removing some D2H syncs in the process.


torch_harmonics/attention/optimized/attention_optimized.py
bonevbs, line 1: why do we need this extra file?

bonevbs (approved):
LGTM, please test distributed tests before merging

PR #217, opened 2026-06-15
Tkurth/16 bit attention

azrael417:
This MR adds FP16 and BF16 dispatching and data handling to the attention layers. It benefits most attention cases by 5-10%, where some edge cases receive a 10% regression. In any case, most cases I tested achieve a slight speedup though.


bonevbs (approved):
LGTM

PR #218, opened 2026-06-15
Tkurth/upsample improvements

azrael417:
This MR fixes performance issues with the upsampling kernel, yielding up to 78x speedup in some cases.


bonevbs (approved):
LGTM

PR #219, opened 2026-06-18
Preserve GaussianRandomFieldS2 sample dimensions

fallintoplace:
## Summary

- Preserve only the trailing singleton distribution dimension when sampling GaussianRandomFieldS2 noise.
- Add regression coverage for single-sample output shape and tiny truncations where `lmax == mmax == 1`.

## Why

`Normal(mean=[0], var=[1]).sample((N, lmax, mmax, 2))` returns a tensor with shape `(N, lmax, mmax, 2, 1)`. The previous broad `squeeze()` could collapse any singleton sample or spectral axis before `view_as_complex`, which breaks sampled noise for valid tiny grids such as `nlat=2` with equiangular or lobatto truncation.

## Validation

- `uv run --no-project --with 'torch>=2.6.0' --with 'numpy>=1.22.4' --with parameterized --with pytest python -m pytest tests/test_random_fields.py -q`
- `uv run --no-project --with ruff ruff check torch_harmonics/random_fields.py tests/test_random_fields.py`


azrael417:
Looks good, thanks for catching this. One request: can you update the Changelog and add this as a comment. Something like: fixing problem with squeezing singleton dims in Noise Tensor for GRF S2.

azrael417 (approved):
LGTM

PR #220, opened 2026-06-18
Fix SpectralConvS2 bias with channel projection

fallintoplace:
## Summary

- Shape `spectral_bias` by input channels in the serial and distributed spectral convolution layers.
- Add regression coverage for `SpectralConvS2(bias=True)` when input and output channel counts differ.

## Why

The spectral bias is applied before the grouped channel contraction, while the tensor is still shaped by input channels. It was previously initialized with `out_channels`, so `bias=True` failed for channel projections such as 4 -> 8 or 8 -> 4.

Keeping the parameter input-channel-shaped preserves the existing pre-contraction bias semantics and keeps equal-channel cases unchanged.

## Validation

- `uv run --no-project --with 'torch>=2.6.0' --with 'numpy>=1.22.4' --with parameterized --with pytest python -m pytest tests/test_spectral_convolution.py -q`
- `uv run --no-project --with ruff ruff check torch_harmonics/spectral_convolution.py torch_harmonics/distributed/distributed_spectral_convolution.py tests/test_spectral_convolution.py`


azrael417 (changes requested):
LGTM, but please amend the Changelog as well here. Thanks for fixing this.

azrael417 (approved):
LGTM.

rietmann-nv (approved):
lgtm

PR #221, opened 2026-06-18
Disable AttentionS2 dropout in eval mode

fallintoplace:
## Summary
- Pass zero dropout to SDPA when `AttentionS2` is in eval mode.
- Add regression coverage for eval determinism with nonzero `drop_rate`, covering both float and tensor scale paths.
- Add a changelog entry.

## Why
`scaled_dot_product_attention` applies dropout whenever `dropout_p` is nonzero; it does not inspect module training state. `AttentionS2` was passing `self.drop_rate` unconditionally, so eval/inference outputs could remain stochastic when dropout was configured.

## Validation
- `uv run --no-project --with 'torch>=2.6.0' --with 'numpy>=1.22.4' --with parameterized --with pytest python -m pytest tests/test_attention.py -q -k 'attention_eval_disables_sdpa_dropout'`
- `uv run --no-project --with ruff ruff check torch_harmonics/attention/attention.py tests/test_attention.py`
- `git diff --check -- torch_harmonics/attention/attention.py tests/test_attention.py Changelog.md`


azrael417 (approved):
LTGM, thanks for fixing this.

PR #222, opened 2026-06-18
fixing the autocasting on disco side

azrael417:
This MR fixes the FP16 and BF16 auto casting in the disco kernel.

ultimately adresses https://github.com/NVIDIA/torch-harmonics/issues/202


bonevbs (approved):
LGTM

PR #223, opened 2026-06-23
DISCO optimizations: forward

azrael417:
This MR adds WGMMA support for DISCO forward (sm_90a support required) and BC tiling for DISCO backward. Both should give a modest performance improvement for sparse psi tensors and a significant one for denser psi tensors.

This MR also adds agent skills for DISCO development and general test writing. I am planning to expand the skills section after this.

This MR should not be merged before the other outstanding MRs are merged, since there are some interdependencies.


tests/test_distributed_convolution.py
```diff
@@ -368,151 +376,95 @@ def test_distributed_disco_conv(
...
+            # small absolute drift near zero-valued weight-gradient entries.
+            pg_atol = max(pg_atol, 3.0)
         if conv_dist.weight.grad is not None:
             wgrad = self._allreduce_param_grad(conv_dist.weight.grad)
             self.assertTrue(compare_tensors("weight grad", conv_local.weight.grad, wgrad, atol=pg_atol, rtol=pg_rtol, verbose=verbose))
         if getattr(conv_dist, "bias", None) is not None and conv_dist.bias.grad is not None:
             bgrad = self._allreduce_param_grad(conv_dist.bias.grad)
             self.assertTrue(compare_tensors("bias grad", conv_local.bias.grad, bgrad, atol=pg_atol, rtol=pg_rtol, verbose=verbose))
 
-    @parameterized.expand(
```
bonevbs, line 378: why is this removed?

tests/test_distributed_convolution.py
```diff
@@ -368,151 +376,95 @@ def test_distributed_disco_conv(
...
+        B, C = 8, 8
 
-        common_args = dict(
+        args = dict(
             in_channels=C,
             out_channels=C,
             in_shape=(nlat, nlon),
             out_shape=(nlat, nlon),
-            basis_type=basis_type,
+            basis_type="piecewise linear",
```
bonevbs, line 411: why do we hardcode this now?

setup.py
```diff
@@ -70,9 +70,30 @@ def get_compile_args(module_name):
...
 
 
+def _torch_cuda_arch_list_has(arch: str) -> bool:
+    """Return True if TORCH_CUDA_ARCH_LIST explicitly requests an arch suffix."""
+    arch_list = os.environ.get("TORCH_CUDA_ARCH_LIST", "").lower()
+    if not arch_list:
+        return False
+
+    normalized = arch_list.replace(";", " ").replace(",", " ")
+    aliases = {
```
bonevbs, line 80: what about earlier architectures? I think we should support sm_80 and above

tests/test_distributed_convolution.py
azrael417: this forces a fallback from packed. PL is the basis function set which is incompatible with kpacked. So it guarantees that the test always will fall back.

setup.py
azrael417: I have not implemented a WGMMA kernel for sm80 yet, only Hopper and Blackwell. Not sure we want to support tensor cores on SM80 since it is an old architecture.

tests/test_distributed_convolution.py
azrael417: this test was only relevant for disco ring, but since we retired ring we do not need it anymore.

setup.py
bonevbs: Ok we can keep this open and use the default kernel for now

bonevbs (approved):
LGTM

PR #234, opened 2026-08-04
Streamline docstrings and fix documentation rendering bugs

azrael417:
Problem
  
  Docstring conventions had drifted across the package, and a few inconsistencies were actually breaking autodoc output. Since the docs have not been published yet, this is a good time to
  normalize them.
  
  Changes
  
  Rendering bugs (visible in the built HTML)
  
  - examples/losses.py: two docstrings (incl. the public forward on the loss base class) were Google-style, but conf.py sets napoleon_google_docstring = False — they rendered as literal prose
  with no parameter table. Converted to NumPy style.
  - examples/models/sfno.py: Example: with a trailing colon isn't recognized by napoleon, so the header and its underline leaked into the page as raw text.
  - 10 API pages published PyTorch's boilerplate ("Define the computation performed at every call") because forward/to lack docstrings and autodoc inherits from nn.Module. Fixed with
  autodoc_inherit_docstrings = False.
  
  Consistency (no rendering change)
  
  - 237 × name: type → name : type; 63 underline lengths corrected; 6 × Example → Examples.
  - Returns unified: single-value → bare type, multi-value → named entries (numpydoc convention).

Docs config
  
  - nitpicky = True with a small nitpick_ignore; with CI's existing -W, a broken cross-reference now fails the build instead of accumulating silently.
  - Added matplotlib to intersphinx_mapping; excluded jupyter_execute (myst-nb copies were rediscovered as source on rebuild).
  - Qualified one :class: reference in distributed_attention.py.
  
  Testing
  
  Full Sphinx build against the working tree: 0 warnings (down from 104 in nitpicky mode). Verified nitpicky is genuinely active by injecting a bogus reference — caught with the new config,
  silent under -D nitpicky=0.
  
  Rendered-output diff through napoleon is exactly 12 lines, all intended Returns changes — confirming the 237 colon and 63 underline edits are render-neutral. AST comparison with docstrings
  stripped shows 0 non-docstring differences across all 27 files.
  
  API / numerical behavior
  
  None. Docstrings, docs/conf.py, and one cross-reference only; no code paths touched.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,10 @@ __pycache__
 *.so
 checkpoints
 
+# Build artifacts
+build/
+*.egg-info
+
 # Kitmaker deployment secrets (never commit)
 scripts/kitmaker.env
 *.env
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -47,7 +47,10 @@ bibtex_default_style = "unsrt"
 bibtex_reference_style = "label"
 
 templates_path = ["_templates"]
-exclude_patterns = ["_build", "Thumbs.db", ".DS_Store"]
+# "jupyter_execute" holds myst-nb's copies of the rendered notebooks; without it
+# a rebuild picks them up as source documents and warns about duplicate labels
+# and pages missing from the toctree.
+exclude_patterns = ["_build", "Thumbs.db", ".DS_Store", "jupyter_execute"]
 
 # MyST (Markdown) options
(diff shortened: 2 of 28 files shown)


albertocarpentieri (commented):
Can we add build/ and *.egg-info to the .gitignore?

albertocarpentieri (approved):
All good, mostly docs and formatting

PR #238, opened 2026-08-10
Derive version from git tag as single source of truth

azrael417:
## Problem
  
  The version was hand-maintained in three places (`__init__.py`, `fallback_version` in
  `pyproject.toml`, `docs/conf.py`) that had to be kept in sync. They drifted, and the
  release bump was missed for `v0.9.2`: since `build_wheels.yml` fed `fallback_version`
  into `SETUPTOOLS_SCM_PRETEND_VERSION`, the tagged build stamped `0.9.2c` — PEP 440
  `0.9.2rc0`, a pre-release `pip install` skips by default. Nothing reached PyPI.

  Two CI bugs contributed: both workflows checked out shallow with `--no-tags` (no tag for
  setuptools-scm to describe against), and the docs job installed with
  `--no-build-isolation` but without `setuptools-scm`, which silently resolves the version
  to `0.0.0` instead of failing — the version the published docs header showed.
  
  ## Change
  
  The git tag becomes the single source of truth. setuptools-scm writes
  `torch_harmonics/_version.py` (generated, gitignored) at build time; `__init__.py` imports
  it, the wheel metadata carries it, and `docs/conf.py` reads it back via
  `importlib.metadata`. `fallback_version` becomes a fixed `0.0.0` sentinel that is no
  longer bumped per release, and the wheel build now fails loudly if it ever resolves to it.
  Both workflows gain `fetch-depth: 0` / `fetch-tags: true`, and the docs job pins the
  version from the release tag.
  
  Releasing is now: update `Changelog.md`, tag, push. No file to bump.
  
  ## Testing
  
  No runtime code paths change. Verified on a clean clone at `v0.9.2` with these changes
  committed and the tag moved onto them (CPU only):
  
  ```bash
  python -c "from setuptools_scm import get_version; print(get_version())"   # -> 0.9.2
  python setup.py egg_info && grep -m1 '^Version:' torch_harmonics.egg-info/PKG-INFO
  # -> Version: 0.9.2      (was 0.9.2rc0)
  
  Also confirmed _version.py ships in sdists via SOURCES.txt, a dirty/off-tag tree
  degrades to 0.9.3.dev0+g<sha> rather than reporting a release version, and Read the Docs
  is unaffected (it installs the package, so importlib.metadata resolves).   
  
  API and behavior changes 
  
  None to the public API. torch_harmonics.__version__ is populated at build time rather
  than hardcoded; unchanged in type and meaning, but reads 0.0.0 in a source tree that was
  never built. The next release publishes as 0.9.2 rather than 0.9.2rc0.

### DIFF
--- a/.github/requirements/build-tools.txt
+++ b/.github/requirements/build-tools.txt
@@ -1,6 +1,10 @@
 # Host-side tooling for the wheel build job (runs on the runner, not in the manylinux container).
 # cibuildwheel drives the build, twine/auditwheel inspect the resulting artifacts.
+# setuptools-scm is needed on the host too, to read the release version off the
+# git tag (see "Determine base version from git"); keep it pinned in lockstep
+# with wheel-build-env.txt so host and container agree on the version.
 cibuildwheel==4.2.0
+setuptools-scm==10.2.1
 twine==7.0.0
 auditwheel==6.7.0
 wheel==0.47.0
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -92,14 +92,68 @@ jobs:
     - name: Determine base version from git
       run: |
         BASE_VERSION=$(python -c "from setuptools_scm import get_version; print(get_version().split('+')[0])")
-        if [ "$BASE_VERSION" = "0.0.0" ]; then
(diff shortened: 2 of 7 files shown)


bonevbs (approved):
lgtm

PR #239, opened 2026-08-10
build CUDA kernels into the released wheels

azrael417:
The cu126/cu128/cu129/cu130 wheels have never contained device code. The stock cibuildwheel manylinux image has no nvcc, no workflow ever set FORCE_CUDA_EXTENSION or its successor, and the auto-detect fallback needs torch.cuda.is_available(), which is false on a GPU-less build runner. Verified against the shipped 0.9.1 and 0.9.2 artifacts: no CUDA symbols, no libcudart, no fatbin. It never failed loudly because the C++ kernels build fine and the layers fall back to torch-native paths, so it survived several releases.

- build the CUDA rows in pytorch/manylinux2_28-builder, pinned by digest
- set TORCH_HARMONICS_BUILD_CUDA_EXTENSION per matrix row
- pass TORCH_CUDA_ARCH_LIST into the build container; it was set on the runner only, so it never reached the build
- use arch-conditional 9.0a/10.0a, which setup.py matches literally to enable the kpacked WGMMA and tcgen05 kernels
- fail the build if a CUDA-labelled wheel contains no device code

### DIFF
--- a/.github/workflows/build_wheels.yml
+++ b/.github/workflows/build_wheels.yml
@@ -347,24 +347,37 @@ jobs:
 
         echo "=== Install + import test ==="
         python -m venv /tmp/test_venv
-        /tmp/test_venv/bin/pip install --quiet torch==${{ matrix.torch-version }} --index-url https://download.pytorch.org/whl/cpu
+        # Install the torch build matching this wheel's CUDA variant, not the CPU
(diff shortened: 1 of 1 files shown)


bonevbs (approved):
LGTM

PR #241, opened 2026-08-14
# Attention: single-conversion NHWC layout contract

azrael417:
## Summary
  
  The attention kernels have always operated on physical NHWC internally, but every
  launcher converted its own inputs on entry and its outputs on exit. This MR makes
  NHWC the *contract* instead: conversion happens once at each module boundary, and no
  kernel converts, and no code infers layout from strides.
  
  The serial path additionally stays in NHWC across the whole module, which turns the
  four 1x1 projections into plain GEMMs and makes qk-norm's head split a free view.
  
  ## What changed

  **Layout is stated, never inferred.** New `torch_harmonics/attention/_layout.py` owns
  both directions (`to_nhwc`/`to_nchw`), backed by the tiled `permute_4D_to0231/0312`
  kernels with autograd and a `permute().contiguous()` fallback when the extension is not
  built. Stride inspection cannot distinguish the two layouts when `C == 1` or `H*W == 1`,
  so the conversion is always explicit.

  **Kernels take NHWC by contract.** All conditional `permute_4D_to0231` blocks are gone
  from the serial and ring launchers; entry points derive extents as `(B, H, W, C)`.
  
  **Heads are packed along channels in the serial op.** `num_heads` is threaded through the
  serial ops; heads are addressed in place rather than folded into batch. This is required
  by the NHWC-throughout module flow, where the projections create the head axis *after*
  the layout conversion, so folding it would materialize a copy.
  
  **Serial module is NHWC end to end.** One `to_nhwc` per distinct input, deduped by
  identity (self-attention costs one conversion, not three); all four 1x1 convs become
  `F.linear`; qk-norm's `(heads, per-head channels)` split is a reshape of contiguous
  memory instead of a 5D permute in and back out; `to_nchw` moved to *after* the output
  projection.
  
  **Ring conversions hoisted out of the loops.** The ring exchange now carries NHWC
  buffers, so received chunks arrive ready to use, and the loop-invariant `qw`/`dy` are
  converted once instead of once per step. Per rank, per fwd+bwd: **11·P + 4 conversions
  -> 14**, independent of ring size.
  
  **Input validation.** Added a dtype-consistency check (`_check_dtypes_match`, plus
  `TORCH_CHECK` in all four serial entry points). The launchers dispatch once on q's scalar
  type and `reinterpret_cast` every activation to it, so a mismatched k/v was previously
  *reinterpreted* rather than converted.
  
  ## Compatibility
  
  No parameter shapes change. Projection weights keep their `(C_out, C_in, 1, 1)` conv
  shape and are viewed to 2D at the call site, so checkpoints load unchanged in both
  directions. `F.conv2d` vs `F.linear` on the same stored weight agrees to 0.0 in fp64. 

## Testing
  
  - `tests/test_attention.py`: 272 passed, 7 skipped (H100 and V100)
  - `tests/test_distributed_attention.py`: all passed 
  - A/B replay of 24 pre-refactor kernel dumps: 216/216 outputs reproduced within the
    recorded run-to-run noise floor, on the machine that captured them
  - New coverage: 14 multi-head AMP rows (fp16/bf16 x downsample/upsample had none
    before), 2 packed-head `opcheck` rows, ring `opcheck` moved to the NHWC ABI
  
  ## Performance
  
  Module-level benchmark on H100: **backward 1.01-1.13x**, forward neutral. The backward
  win is the expected shape — it previously re-converted k/v/q and dy on entry and three
  gradients on exit, and now reuses the forward's converted tensors. The serial forward
  calls its launcher once, so moving the conversion to the boundary removes no work there;
  the ring is where a forward win exists, and it is **not yet measured**.

 ## Deliberately not done
  
  - `AttentionS2` stays on `conv2d`/NCHW. It is the reference in
    `test_neighborhood_global_equivalence`, which loads its `state_dict` into
    `NeighborhoodAttentionS2` — every other correctness test now has the reworked 
    projection code on both sides, so migrating it would blind the only independent check.
  - CPU kernels keep the whole-tensor fp32 upcast. They are hardcoded fp32 accessors, and
    CPU reduced-precision arithmetic is emulated, so there is no bandwidth to win.


bonevbs (approved):
LGTM

PR #242, opened 2026-08-15
Radial Extension for SFNO

tmarkmann:
# Summary
This PR builds towards an extension of the current SFNO to 3D systems including a radial domain. Planned additions are new PDE examples, a Mellin Convolution and a new Radial Spherical Neural Operator Model. For now this PR only includes an addition of the spherical Poisson Equation for the half-line and exterior domain. More updates will be added to the PR in the future.

# Roadmap
- [x] Poisson Equation
- [ ]  Mellin Convolution
- [ ]  Radial SFNO Model

# Main Additions and Changes
`torch_harmonics/examples/poisson_equation.py`: Added Poisson Solver
`torch_harmonics/examples/poisson_dataset.py`: Added Dataset for training
`torch_harmonics/quadrature.py`: Added geometric weights
`notebooks/poisson_equation.ipynb`: Examples for Poisson


tmarkmann:
@bonevbs

azrael417:
Hello Thorben — nice work, the solver is mathematically sound. Thanks for the MR. With the help of Claude, I pushed some changes to streamline it with the rest of the repo:

I validated the Green's operator against analytic solutions before touching anything: the monopole case u = exp(-ar²) converges cleanly at second order (rel L∞ 2.97e-4 → 7.39e-5 → 1.85e-5 for nr = 256/512/1024), the l=1 case lands at 3.2e-4, and the exterior harmonic lift is exact to 6.6e-15. The exterior image term makes the kernel vanish identically at r = R, so the homogeneous Dirichlet condition is exact rather than approximate.

I pushed five commits directly to the branch — happy to revert any of them if you disagree. Summary:

1. Removed the einops dependency (45a64bf)

This was the one blocker. einops isn't in pyproject.toml and isn't used anywhere else in the repo, and since examples/__init__.py imports poisson_equation eagerly, it broke every existing consumer of torch_harmonics.examples for anyone without it installed — the shallow water / depth / segmentation training scripts, the helmholtz and SFNO notebooks, Dockerfile.examples. CI wouldn't have caught it, because the test suite never imports examples.

All nine rearrange calls were pure axis insertions (no permutes), so they became unsqueeze/reshape. Verified bit-for-bit identical: the green and blift buffers, random_source, and solve all compare equal on both domains.

2. Reused precompute_latitudes (a19f51b)

The grid if/elif was copied from pde_sphere.py; the utility already does this and handles the unknown-grid case. lats = π/2 - colats reproduces the previous -arcsin(cost) values to ≤3.3e-16 on all three grids with identical ordering, and lats only feeds the plotting helpers, so solutions are untouched.

3. Removed currently-unreachable code (eb8e63e)

Since the Mellin convolution and radial SFNO are still ahead, these seemed better added alongside the work that needs them:

- radial_grid's "shell" branch — GreensOperator rejects any domain but half-line/exterior, and the solver passes its own domain straight through, so nothing could reach it.
- solve_density and the gravity buffer — no callers; it's a constant prefactor on solve().
- plot_sphere's "orthographic" branch — the notebook only uses "mollweide", and this was the sole cartopy reference (also not a declared dependency).

Same commit documents geometric_weights in docs/api/utilities.md, notes the dense (lmax, nr, nr) buffer size (33 MB at nlat=64/nr=256, but 537 MB at nlat=256/nr=512, and PoissonDataset runs one solve per sample), and corrects the scale docstring — it uses an unweighted angular mean, not the sphere measure.

4. ValueError instead of assert in plot_sphere (12f9e1b)

torch_harmonics/ has had zero bare asserts since the v0.9.2 conversion, so this was the only one; matches plotting.py's style.

5. Wired the notebook into the docs (40e2275)

All 13 other notebooks are symlinked into docs/tutorials/ and in the toctree; this was the only one missing. Filed under Applications next to helmholtz. Confirmed with a local sphinx -b html . build -W --keep-going that the tutorial page and the geometric_weights stub both render warning-free.

azrael417 (dismissed):
LGTM now, I applied some changes. Please have a look if they make sense. Then I recommend rebasing your code onto main once this MR lands.

tmarkmann:
Hey Thorsten, thank you for your review! All of your points are fine for me and I will adapt future code to be more in line with the lib. Here are my thoughts on your points:

1. I think einops makes the code much more understandable but if it's not used throughout the library I will switch to reshapes from now to not break the library.

2. I copied that from examples/shallow_water_equations.py (and pde_sphere.py) so maybe it should be changed there as well.

3. I'm using some of the unused code in a downstream repository but I can readd those when they are actually used from torch-harmonics. One note about the cartopy plotting reference: I basically copied that from the examples/shallow_water_equations.py as well. Since it is a lazy import I think that should be fine and I can add an orthographic plot to the notebook (so it is called already).

One general question from my side is whether we wait until all the features are implemented and pushed to this PR or merge the smaller bits. I'm fine with both but @bonevbs might have a preference from earlier discussions.

azrael417:
Hello Thorben, thanks for the quick reply. We are trying to keep the numbers of package dependencies down as much as possible. Einops can be OK as long as it is contained in the examples, since those have their own set of dependences anyway, but most of the ops used were just permutes and reshapes, which I prefer doing in torch directly. Afair, I have removed all the einops from torch-harmonics otherwise, please let me know if that is not correct. 

I have removed some of the unused code so that it can be added later in another targeted MR. So this MR adds the poisson equation and solver and some plumbing and a notebook. The next can then add the additional features like the GreensFunction and other classes. Does that make sense? Claude suggested to split it there. We will certainly add it back when you make a follow-up MR. This is part 1 of 3, right?

torch_harmonics/quadrature.py
```diff
@@ -171,6 +171,37 @@ def trapezoidal_weights(n: int, a: Optional[float] = -1.0, b: Optional[float] =
     return xlg, wlg
 
 
+def geometric_weights(n: int, a: Optional[float] = 1.0, b: Optional[float] = math.e) -> Tuple[torch.Tensor, torch.Tensor]:
```
bonevbs, line 174: can we rename this to radial weights? maybe even better to name them 'log_uniform_weights' ? given that they are uniform in logarithmic domain

bonevbs: also is there a good reason to name the parameters a and b? I would potentially name them inner_radius and log_base

torch_harmonics/examples/poisson_dataset.py
```diff
@@ -0,0 +1,133 @@
...
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
+
+
+import torch
+
+from .poisson_equation import RadialPoissonSolver
+
+
+class PoissonDataset(torch.utils.data.Dataset):
```
bonevbs, line 38: torch-harmonics has a PDEDataset object. Couldn't we have reused that? or derived from that?

bonevbs (commented):
Hi @tmarkmann most of this already looks good, thank you! I left some comments. The bigger one is of this couldn't just reuse the PDEDataset as this would make it easier to just reuse existing training pipelines in the examples folder.

bonevbs:
For the einops, while I do agree it might make it more readable, we are usually trying to keep dependencies lightweight. So I prefer removing it unless it becomes absolutely essential.

torch_harmonics/quadrature.py
tmarkmann: I agree that 'log_uniform_weights' is a much better name. The parameters were called a and b to mirror 'trapezoidal_weights'. I think there are two options:
1.  'log_uniform_weights' and stick to a and b
2. 'radial_weights' and change to inner_radius and log_base

I would go for the first one if we ever add a shell as an option. I suppose that we could have uniform radial weights in that case and this could lead to confusion if we already reserve 'radial_weights'.

torch_harmonics/examples/poisson_dataset.py
tmarkmann: I did not extend the PDEDataset because I did not want to break existing code and some things are different for the poisson equation:
1. radial dimension: we need to either have a 2D or 3D dims object depending on the pde
2. domain: half-line or exterior which only is important for poisson
3. initial_condition: there is only random for poisson
4. the normalization for poisson needs to be separate for target and input (different to swe)

My first thought was that caring about all the divergences between SWE and Poisson is a bit hacky. But for the example poisson I can just fix all parameters like domain and rmin, rmax and then it should be ok. I will fix that!

tmarkmann:
> So this MR adds the poisson equation and solver and some plumbing and a notebook. The next can then add the additional features like the GreensFunction and other classes. Does that make sense? Claude suggested to split it there. We will certainly add it back when you make a follow-up MR. This is part 1 of 3, right?

Yes, this is part 1 of 3. Since the other two parts will have larger changes I agree to split it here. I will implement @bonevbs changes later today and then it is ready from my side.

torch_harmonics/quadrature.py
azrael417: I am fine with 1, that sounds good. But make sure the doctoring describes what the parameters do. If we change the naming of the parameters here, we also should do that for the trapezoidal weights imo. Boris, what would be good names there? Then we can fix these issues in one go.

torch_harmonics/examples/poisson_dataset.py
azrael417: Maybe we should rename the PDE dataset to SWEDataset? Since that is more descriptive? Boris, what do you think?

azrael417: I do not mind having a separate class per example, in that way each example is self contained, up to the shared content from torch harmonics that is. But users can grab the example and their TH install and run it, no need to use classes or functions from an example helper folder or something like that.

PR #243, opened 2026-08-17
Return float64 weights from trapezoidal_weights

azrael417:
## Problem

`trapezoidal_weights` builds its weights with `torch.ones(n)`, which picks up torch's default dtype. The weights came back **float32** while the nodes, built from `np.linspace`, are **float64**.

It is the only rule in the module doing this:

| Function | Nodes | Weights |
|---|---|---|
| `legendre_gauss_weights` | float64 | float64 |
| `lobatto_weights` | float64 | float64 |
| `clenshaw_curtiss_weights` | float64 | float64 |
| `trapezoidal_weights` | float64 | **float32** |

The visible public path is `precompute_latitudes(grid="equiangular-trapezoidal")`, whose weights were float32 while the other three grids were float64.

The mismatch can also hide downstream: multiplying float64 nodes by float32 weights *promotes* back to float64, so a derived rule can return a float64 tensor that already carries float32 rounding. A dtype check on the output would not catch it — only a precision test does.

## Fix

```python
wlg = (b - a) / (n - 1 + periodic * 1) * torch.ones(n, dtype=xlg.dtype, requires_grad=False)
```

## Impact

`QuadratureS2` integrating `f = 1` over the sphere in float64:

| Grid | Before | After |
|---|---|---|
| `equiangular-trapezoidal` | 8.4e-07 | **9.7e-08** |
| `equiangular` (reference) | 8.2e-08 | 8.2e-08 |

The trapezoidal grid now matches the others. The residual is the layer casting its `quad_weight` buffer to float32, which this PR does not touch.

## Tests

- **New `TestQuadratureWeightPrecision`**: compares the weights against a float64 analytic reference at `1e-15` (the uniform weight `(b-a)/(n-1)` is not exactly representable, so float32 leaves a ~1e-8 relative error), and asserts node/weight dtype consistency across all four grids. **These fail on `main` and pass here** — 4 failures without the fix.
- **Tightened** the two `equiangular-trapezoidal` cases in `test_constant_integral` from `1e-4` to the `1e-6` used by the other grids, since the measured error is now identical to equiangular's. Note this tightening alone does *not* catch the bug — the `rtol` term leaves enough slack either way — so it is a consistency cleanup, not the regression test.
- **Left `test_polynomial_latitude_integral` at `1e-2`** for the trapezoidal grid: that error is the genuine O(h²) of the rule, measured at 2.1e-03, not a precision artifact.

Full non-distributed suite run against both versions: **391 passed, 16 skipped**, identical either way.

## Note on merge order

This opens a `### v0.9.3` changelog section, as does #242. Whichever merges second will hit a trivial conflict on the heading line.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,10 @@
 
 ## Versioning
 
+### v0.9.3
+
+* Fixed `trapezoidal_weights` returning float32 weights alongside float64 nodes, because the underlying `torch.ones(n)` inherited the default dtype. It was the only rule in `torch_harmonics.quadrature` doing so, and it capped the accuracy of everything derived from it at roughly 1e-7, including the latitude weights of the `"equiangular-trapezoidal"` grid. Weights are now float64 like the other rules, and the tolerances of the affected tests have been tightened to match the other grids.
+
 ### v0.9.2
 
 * Added upsampling support to `DistributedNeighborhoodAttentionS2` (`nlon_out % nlon_in == 0`): new upsample (scatter) ring-step CUDA kernels for forward and backward, matching the serial upsample attention. K/V rotate around the azimuth ring while queries and the softmax state stay local; all three directions (self-attention, downsampling, upsampling) are now supported by the distributed layer.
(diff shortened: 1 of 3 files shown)


bonevbs (approved):
LGTM

PR #244, opened 2026-08-18
Attention fwd: neighbor-groups + segment-based psi iteration (remove per-neighbor integer division)

azrael417:
Summary

The optimized CUDA forward kernel spent most of its time on address arithmetic rather than math: profiling on H100 showed 80% compute throughput while delivering only ~2.4% of peak FLOPs, DRAM at 0.6%, and 9.9 cycles/issued instruction — i.e. instruction-bound on a per-neighbor 64-bit integer division (recovering the input latitude hi from col_idx), which the GPU emulates in ~70–100 instructions against ~4 instructions of useful math per neighbor.

This PR removes that division from the hot loop in two steps:

1. Segment-based psi. psi's sparsity is a union of contiguous longitude arcs — one per (output row, input latitude) — which follows geometrically from a geodesic ball intersecting a latitude circle in a single arc. Adds _build_psi_segments/_expand_psi_segments to re-express the column list as (hi, lo, len) arcs, plus TestPsiArcStructure to pin the contiguity property and a round-trip test (expand-segments-back-to-col_idx) that's strictly stronger than a contiguity check alone (catches off-by-one arc starts that contiguity alone would miss). Verified exact across equiangular/legendre-gauss/lobatto grids, gather and scatter, pscale 1/2, odd latitude counts, and a wide-cutoff case that exercises wrapping arcs.
2. Kernel consumption. Both CUDA forward kernels now walk psi as arcs: hi and the quadrature weight become per-arc constants, and the column advances by counting instead of dividing. psi_seg/psi_seg_off are added to the forward op's ABI alongside the existing col_idx/row_off, which are kept because the CPU and torch reference paths still consume them — deliberately, so test_custom_implementation keeps comparing optimized-on-segments against torch-on-col_idx, and any segment-construction bug shows up as a disagreement rather than hiding behind a shared bug.

Separately, the special forward kernel now processes neighbors in groups of 4 (NB=4): four col_idx/address computations and k/v loads are hoisted per step instead of one, and online softmax is applied once per group rather than once per neighbor (same reduction algebraically — one running max, one rescale per group). Numerics move toward the fp64 reference, not away, but are not bit-identical to before in reduced precision.

Not in scope: the scatter/upsample kernels have the same per-neighbor division (attention_cuda_fwd_upsample.cu:150, :207, and three backward sites) but no benchmark config currently exercises them, so they're left unmeasured/unchanged for now — flagged as follow-up.

Testing

- tests/test_attention.py — 290 passed / 7 skipped on both V100 and H100, plus the distributed suite (per commit c075282).
- tests/test_attention.py -k custom_implementation — 174 passed (CPU + CUDA, fp32/fp16/bf16, multi-head, both resampling directions) for the neighbor-groups change (per commit bd61199).
- New tests: TestPsiArcStructure (contiguity) and the segment round-trip test added in 2d9bb6f/3da7658.

Performance

Op-level forward timing (H100, attention_kernels::forward) for the neighbor-groups change:

| Config | Before | After | Speedup |
|---|---|---|---|
| 1deg_tc0017 | 0.272 ms | 0.180 ms | 1.51x |
| 1deg_tc003 | 0.542 ms | 0.447 ms | 1.21x |
| hdeg_tc0017 | 2.190 ms | 2.058 ms | 1.06x |
| hdeg_tc003 | 6.632 ms | 6.148 ms | 1.08x |

Win shrinks with problem size — at scale the kernel is memory-request-bound (no cross-query reuse, ~256 B moved per query/neighbor pair vs. a tiled kernel's ~1), a separate and larger problem not addressed here.

### DIFF
--- a/tests/test_attention.py
+++ b/tests/test_attention.py
@@ -398,7 +398,9 @@ class TestNeighborhoodAttentionS2(unittest.TestCase):
         qw = torch.randn(2, nlat, nlon, channels, device=self.device, dtype=torch.float32)
 
         with torch.autocast(self.device.type, dtype=autocast_dtype):
-            out = torch.ops.attention_kernels._neighborhood_s2_attention_optimized(kw, vw, qw, model.quad_weights, model.psi_col_idx, model.psi_roff_idx, 1, nlon, nlat, nlon)
+            out = torch.ops.attention_kernels._neighborhood_s2_attention_optimized(
+                kw, vw, qw, model.quad_weights, model.psi_col_idx, model.psi_roff_idx, model.psi_seg, model.psi_seg_off, 1, nlon, nlat, nlon
+            )
 
         self.assertEqual(out.dtype, autocast_dtype, f"autocast output dtype {out.dtype} != {autocast_dtype}")
 
--- a/tests/test_attention.py
+++ b/tests/test_attention.py
@@ -43,6 +43,7 @@ from torch.library import opcheck
 # from torch.autograd import gradcheck
 from torch_harmonics import AttentionS2, NeighborhoodAttentionS2
 from torch_harmonics.attention import cuda_kernels_is_available, optimized_kernels_is_available
+from torch_harmonics.attention._attention_utils import _build_psi_segments, _expand_psi_segments
 from torch_harmonics.attention._layout import to_nhwc
 from torch_harmonics.attention.kernels_torch.attention_torch import (
     _neighborhood_s2_attention_bwd_dk_torch,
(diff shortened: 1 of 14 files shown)


rietmann-nv (approved):
Seems fine. Definitely a good catch reducing expensive integer instructions.

mauro-bis:
Seems good to me. Fwd/bwd kernels spend a significant amount of time on indexing arithmetic so expressing the computations in a way that allows to reduce/eliminate expensive instructions involved (like division and modulo) is a good idea.

PR #245, opened 2026-08-19
Attention bwd: arc-segment iteration + upsample backward optimization

azrael417:
Base: tkurth/attention-fwd-neighbor-groups (#244) — stacked on top; do not merge before #244.

Summary

Follow-up to #244, which removed the per-neighbor 64-bit integer division from the forward kernels by walking psi as contiguous longitude arcs instead of decoding a flat column index. This PR applies the same idea to the backward kernels, plus a first pass at the previously-unoptimized upsample (scatter) backward.

1. Gather backward: arc segments. All four neighbor loops (generic and special kernels, two passes each) now derive (input lat, input lon) by counting along an arc instead of decoding col_idx — same division/modulus removal as the forward. col_idx/row_off are dropped from the kernel signatures; row_off survives only in the dispatcher, which still uses row lengths for sortRows load balancing. The op schema keeps both col_idx and the segments, since the CPU backward and the torch reference still walk the column list — keeping the reference independent of the arc derivation, same rationale as the forward PR.

1. Measured on V100, fp32, C=64, backward only:

| Config            | Before    | After     | Δ    |
|-------------------|-----------|-----------|------|
| 180x360 self      | 39.68 ms  | 37.50 ms  | 5.5% |
| 360x720 → 180x360 | 159.06 ms | 150.02 ms | 5.7% |

1. Much smaller than the forward's 2.02x: the forward was instruction-bound on address arithmetic, the backward isn't, so the division was a small share of its cost. Kept because it's free and simplifies the loops — the backward's real bottleneck (atomicAdd traffic in the gradient pass, four __warp_sum reductions per neighbor) is still unprofiled.
3. Special kernel: qy/dy in registers. The special kernel re-read sh_qy/sh_dy from shared memory once per neighbor despite both being loop-invariant per-thread slots. ncu showed it issue-limited (69% peak LSU issue, 84% SM throughput, 1.1% DRAM, "not selected" the largest stall). Per-thread reads move to registers; the shared copies stay because the scalar epilogue reads across the warp and needs them.

2. Measured on V100, fp32, C=64, backward only:

| Config            | Before    | After     | Δ    |
|-------------------|-----------|-----------|------|
| 180x360 self      | 37.50 ms  | 36.54 ms  | 2.6% |
| 360x720 → 180x360 | 150.02 ms | 146.37 ms | 2.4% |

4. Upsample (scatter) backward: arc segments + sortRows. The upsample backward's three neighbor-walking kernels (max_k, stats_k, dkv_k) are structurally a gather despite the "scatter" naming — ncu confirms 0.00% L2 atomic cycles, since each kernel parallelizes over input points that already own their output. So the gather-side optimizations apply and, per the file's own prior comment ("no specialized channel-size variant or sortRows"), none had been applied yet. This PR lands two of the four: arc segments (removing the division) in all three loops, and sortRows (rows visited longest-first — psi's density is heavily latitude-skewed, and natural order left entire latitudes straggling; ncu measured 29.95% achieved occupancy against 50% theoretical).

3. Measured on V100, fp32, C=64, 180x360 → 360x720 backward:

| Kernel          | Before   | After            |
|-----------------|----------|------------------|
| scatter_dkv_k   | 152.3 ms | 108.6 ms         |
| scatter_stats_k | 127.7 ms | 92.1 ms          |
| scatter_max_k   | 65.5 ms  | 34.6 ms          |
| total           | 349.4 ms | 241.2 ms (1.45x) |

3. Far more than the 5% the segment change alone bought the gather backward — this path is latency-bound rather than issue-bound. The two changes (segments + sortRows) were applied together, so their individual contributions aren't separated.

3. Not done here: specialized channel-size variant and vectorization for the upsample backward. ncu shows 64 registers/thread capping occupancy at 50% theoretical, now the binding constraint on dkv_k — flagged as follow-up, same as the scatter/upsample forward division was flagged as follow-up in #244.
5. Benchmarks. Adds quarter-degree (721×1440, ~1.04M query points, 16x the 1° entries) self-attention configs (fp32/fp16/bf16) to benchmarks/attention.py — the resolution production is actually headed for, where kernels stop being dominated by launch/latency effects. Correctness checks are skipped for these: the dense torch reference is quadratic in nlat*nlon and doesn't fit at this size.

Testing

- Suite: 302 passed / 7 skipped on H100.
- Op schema change (seg/seg_off now saved in ctx.save_for_backward alongside col_idx/row_off) is covered by the existing opcheck-based PT2 compatibility tests and test_custom_implementation.

API / numerical changes

- Backward custom op's saved-context tuple grows from (col_idx, row_off, quad_weights, kw, vw, qw) to (col_idx, row_off, seg, seg_off, quad_weights, kw, vw, qw) — positional, and _neighborhood_s2_attention_bwd_torch's unpack was updated to match (segments unused there, kept only to share _setup_context_attention_backward with the CUDA path). No public Python API change.
- No numerical changes: backward arc-segment iteration is a pure re-derivation of the same column indices, and sortRows/register changes don't alter reduction order in a way that should move results outside existing tolerances (call this out explicitly in the PR if you saw any tolerance adjustments needed — I didn't see one in the diff, but flag if I'm missing something).

### DIFF
--- a/.github/requirements/build-tools.txt
+++ b/.github/requirements/build-tools.txt
@@ -6,5 +6,5 @@
 cibuildwheel==4.2.0
 setuptools-scm==10.2.1
 twine==7.0.0
-auditwheel==6.7.0
-wheel==0.47.0
+auditwheel==6.8.1
+wheel==0.48.0
--- a/.github/requirements/style-tools.txt
+++ b/.github/requirements/style-tools.txt
@@ -1,3 +1,3 @@
 # Linting tooling, pinned so that style runs are reproducible and do not change under us.
 # The individual hook versions are pinned separately in .pre-commit-config.yaml.
-pre-commit==4.6.1
+pre-commit==4.6.2
--- a/.github/requirements/wheel-build-env.txt
+++ b/.github/requirements/wheel-build-env.txt
@@ -3,6 +3,6 @@
 # extension and stamps the version into the published wheels. Pinned so that a compromised or
 # simply broken upstream release cannot silently change what we ship.
 pip==26.2.1
-setuptools==83.0.0
-wheel==0.47.0
+setuptools==84.0.0
+wheel==0.48.0
 setuptools-scm==10.2.1
--- a/benchmarks/attention.py
+++ b/benchmarks/attention.py
@@ -578,6 +578,59 @@ _NATTN_CONFIGS = [
         skip_correctness=True,
         tags=["attention", "neighborhood", "cross"],
     ),
+    # quarter-degree self-attention. The resolution the production configs are
+    # actually headed for, and the one where the kernels stop being latency-bound on
+    # grid size alone: 721x1440 is ~1.04M query points, 16x the 1deg entries.
+    # Correctness is skipped -- the dense torch reference is quadratic in nlat*nlon
+    # and does not fit at this size.
+    dict(
+        name="nattn_s2_opt_qdeg_b1_c64_h1_tc003_float32_cuda",
+        device="cuda",
+        dtype=torch.float32,
+        batch=1,
+        channels=64,
+        num_heads=1,
+        nlat_in=721,
+        nlon_in=1440,
+        nlat_out=721,
+        nlon_out=1440,
+        theta_cutoff=0.03,
+        optimized=True,
+        skip_correctness=True,
+        tags=["attention", "neighborhood", "self"],
+    ),
+    dict(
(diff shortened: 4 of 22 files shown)


mauro-bis:
Everything looks good to me. Green light for the merge from my side.

azrael417:
Great, thanks for your help Mauro.

bonevbs (approved):
LGTM

PR #246, opened 2026-08-19
Register AutocastCPU for the attention and DISCO custom ops

azrael417:
Both op families registered an autocast kernel only at the AutocastCUDA dispatch key, so under torch.autocast("cpu", ...) nothing reconciled their inputs. The two symptoms differ:

- Attention hard-fails. The kernels dispatch once on q's scalar type and then reinterpret every activation pointer as that type, so they require k, v and q to share a dtype and check it explicitly. Autocast casts some ops and not others, so a module mixing projections with normalization can hand the op an fp32 q next to an fp16 v, tripping the check: RuntimeError: v dtype (Half) must match q dtype (Float)

- DISCO fails quietly. Its CPU kernel normalizes dtypes itself (disco_cpu_fwd.cpp upcasts reduced-precision inp and vals to fp32), so nothing errors -- but the op ran at whatever dtype it was handed rather than the autocast dtype, i.e. CPU autocast simply had no effect on DISCO layers.

Both now register AutocastCPU alongside AutocastCUDA, via one closure per op so the two keys cannot drift apart again.

Found on nvcr.io/nvidia/pytorch:24.12-py3 (torch ~2.6), where six CPU AMP rows of test_custom_implementation failed. It was invisible on newer PyTorch, on CI and on two other containers, because those happen to produce consistent dtypes for this graph -- the missing registration was latent there, not absent.

The new test is therefore at op level rather than module level: it constructs the mismatch directly (fp32 q, reduced-precision k/v) so the regression is detectable on any torch version. Without autocast active the same inputs raise, which makes the test discriminating rather than vacuous.

Suite: 596 passed, 15 skipped (attention + convolution).

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -12,6 +12,7 @@
 * Added fused distributed DISCO convolution variant which reduces activation storage.
 * Adding WGMMA (tensor core) support to DISCO forward kernels on H100 (SM90) architectures. The kernel will be selected automatically if shapes allow, no specific action from the user is required.
 * Improved performance for CPU based attention kernels.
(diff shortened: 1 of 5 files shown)


bonevbs (approved):
LGTM

PR #250, opened 2026-08-20
Unify the theta_cutoff heuristic and derive it from the grid

azrael417:
### Problem
     
  The default angular cutoff for DISCO convolutions and neighborhood attention was
  computed as `pi / (nlat - 1)` in six independent places:
     
  - `disco/convolution.py:564` (forward), `:926` (transpose)
  - `distributed/distributed_convolution.py:243` (forward), `:524` (transpose)
  - `attention/attention.py:355`, `:357`
     
  That expression is the exact latitudinal node spacing of an **equiangular**
  (Clenshaw-Curtis) grid, but not of the others:
     
  | grid | `pi/(nlat-1)` | actual max spacing | ratio |
  |---|---|---|---|
  | equiangular | 0.049087 | 0.049087 | 1.00 |
  | legendre-gauss | 0.049087 | 0.047962 | 0.98 |
  | lobatto | 0.049087 | 0.059408 | **1.21** |
  | equiangular-trapezoidal | 0.049087 | 0.250656 | **5.11** |
  
  (nlat = 65, same grid in/out.)
  
  The root cause of the last row is that `"equiangular-trapezoidal"` is a misnomer:
  `precompute_latitudes` builds it via `trapezoidal_weights` on the `cos(theta)`
  interval `[-1, 1]`, so its nodes are equispaced in `cos(theta)`, not in `theta`.
  
  The consequence is a silently degenerate polar stencil. Since the great-circle
  distance between `(theta_o, 0)` and `(theta_i, phi)` is bounded below by
  `|theta_o - theta_i|`, a latitude ring outside the cutoff in `theta` cannot
  contribute to the psi row for any `phi`. Counting rings within the old default:

| grid | `pi/(nlat-1)` | actual max spacing | ratio |
  |---|---|---|---|
  | equiangular | 0.049087 | 0.049087 | 1.00 |
  | legendre-gauss | 0.049087 | 0.047962 | 0.98 |
  | lobatto | 0.049087 | 0.059408 | **1.21** |
  | equiangular-trapezoidal | 0.049087 | 0.250656 | **5.11** |
     
  (nlat = 65, same grid in/out.)
     
  The root cause of the last row is that `"equiangular-trapezoidal"` is a misnomer:
  `precompute_latitudes` builds it via `trapezoidal_weights` on the `cos(theta)`
  interval `[-1, 1]`, so its nodes are equispaced in `cos(theta)`, not in `theta`.
     
  The consequence is a silently degenerate polar stencil. Since the great-circle
  distance between `(theta_o, 0)` and `(theta_i, phi)` is bounded below by
  `|theta_o - theta_i|`, a latitude ring outside the cutoff in `theta` cannot
  contribute to the psi row for any `phi`. Counting rings within the old default:
  
  | grid | nlat=33 | nlat=65 | nlat=129 |
  |---|---|---|---|
  | equiangular | 0 | 0 | 0 |
  | legendre-gauss | 0 | 0 | 0 |
  | lobatto | 2 | 4 | 6 |
  | equiangular-trapezoidal | 8 | 14 | **30** | 
  
  (output latitudes whose neighborhood collapses to the single ring they sit on.)
  
  ### Change
  
  All six sites now call `torch_harmonics.quadrature.compute_theta_cutoff`, backed by
  `compute_latitude_spacing`, which takes one latitudinal grid spacing from the grid's
  actual node distribution. Which grid sets the cutoff is unchanged at every call site:
  output grid for a forward transform, input grid for a transpose one.
  
  This is the single place the heuristic now lives, and it is written to take a grid
  descriptor once the descriptor-based API lands, at which point `nlat` and `grid`
  collapse into one argument.

### Numerical behavior change
  
  **Equiangular is unaffected in practice.** Its cutoff moves by ~1e-15 (the new value
  comes back through `arccos` of the Clenshaw-Curtis nodes), and the resulting psi has a
  bit-identical sparsity pattern with values agreeing to 1e-13 in float64 — far below
  float32 resolution. `TestThetaCutoffPsiRegression` asserts exactly this.
  
  Other grids do change: lobatto widens ~21%, equiangular-trapezoidal ~5x, legendre-gauss
  narrows ~2.3%. Legendre-gauss narrowing is deliberate — the node spacing is the
  definition we want, not a floor on the old expression. Models trained on non-equiangular
  grids see a different stencil unless `theta_cutoff` is passed explicitly.
  
  Following the `truncate_sht` precedent from v0.9.0, a `UserWarning` is emitted on grids
  whose default changed. Equiangular does not warn. The message is direction-aware, since
  legendre-gauss narrows rather than under-covering.

### Tests
  
  New `tests/test_grid_assumptions.py` pins the invariants this relies on: the cutoff
  covers each grid's maximum node spacing and yields a multi-ring neighborhood; equiangular
  psi is unchanged; where each grid's nodes are equispaced; and the pole symmetry of nodes
  and weights — which is what currently makes the differing flip conventions of
  `precompute_latitudes` and `QuadratureS2` (`quadrature.py:403`) agree by accident.
  
  `test_convolution.py::test_convolution_tensor_integrity` computed its own
  `pi / (nlat_out - 1)` and passed it explicitly, so it pinned the old behavior instead of
  exercising the module default — including on its legendre-gauss rows. It now uses
  `compute_theta_cutoff`.
  
  Added lobatto and equiangular-trapezoidal rows to both the serial and distributed conv
  parameterizations. Neither grid appeared in any conv test before, so the grids whose
  default actually changed had no module-level coverage.
  
  **Test runs** (H100, NGC 26.04 / torch 2.12.0a0):
  
  | suite | result |
  |---|---|
  | full serial suite | 1590 passed, 18 skipped |
  | `test_distributed_attention.py`, 8 ranks | 68 passed × 8 |
  | `test_distributed_convolution.py`, 8 ranks | 48 passed × 8 |
  
  The distributed runs are on a real multi-rank split, not 1×1, so the new 
  non-equiangular rows exercise the polar halo and split bookkeeping.

### Performance note
  
  Not a kernel rewrite, so no benchmark table. One consequence worth flagging: on
  `"equiangular-trapezoidal"` the ~5x wider cutoff makes psi substantially denser, so
  DISCO and neighborhood attention on that grid get correspondingly slower and use more 
  memory. That is the cost of covering the poles it was previously missing. Given the name
  is misleading for a `cos(theta)`-equispaced grid, deprecating it may be worth a separate
  discussion.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,8 +2,10 @@
 
 ## Versioning
 
-### v0.9.3
+### v0.9.3b1 (unreleased)
 
+* Unified the default `theta_cutoff` heuristic for DISCO convolutions and neighborhood attention into `torch_harmonics.quadrature.compute_theta_cutoff`, backed by the new `compute_latitude_spacing`. The cutoff was previously computed as `pi / (nlat - 1)` in six independent places; that expression is the exact latitudinal node spacing of an equiangular grid, but not of the others. Gauss-Lobatto nodes cluster towards the equator (polar spacing ~21% larger) and `"equiangular-trapezoidal"` nodes are equispaced in `cos(theta)` rather than in `theta` (polar spacing ~5x larger), so on both grids the old default under-covered the poles: at `nlat = 129`, 6 of 129 output latitudes on a lobatto grid and 30 of 129 on an equiangular-trapezoidal grid saw only the single latitude ring they sit on, silently degenerating the stencil there. The cutoff is now taken from the grid's actual node distribution.
+* **Breaking**: the default `theta_cutoff` of `DiscreteContinuousConvS2`, `DiscreteContinuousConvTransposeS2`, `NeighborhoodAttentionS2`, and their distributed counterparts now depends on the grid's node distribution rather than on `nlat` alone. Equiangular grids, which are the default, are unaffected in practice — the value moves by ~1e-15 and the resulting `psi` has an identical sparsity pattern — but `"lobatto"` widens by ~21%, `"equiangular-trapezoidal"` by ~5x, and `"legendre-gauss"` narrows by ~2.3%. Models trained on non-equiangular grids will see a different stencil unless `theta_cutoff` is passed explicitly. A `UserWarning` is emitted on the grids whose default changed, following the `truncate_sht` precedent from v0.9.0.
 * Fixed `trapezoidal_weights` returning float32 weights alongside float64 nodes, because the underlying `torch.ones(n)` inherited the default dtype. It was the only rule in `torch_harmonics.quadrature` doing so, and it capped the accuracy of everything derived from it at roughly 1e-7, including the latitude weights of the `"equiangular-trapezoidal"` grid. Weights are now float64 like the other rules, and the tolerances of the affected tests have been tightened to match the other grids.
 
 ### v0.9.2
--- a/docs/api/utilities.md
+++ b/docs/api/utilities.md
@@ -8,15 +8,20 @@ torch-harmonics supports several quadrature rules for the latitudinal
 direction. Each corresponds to a `grid` keyword accepted by the SHT and
 convolution layers:
 
-| Grid string                 | Quadrature rule | Nodes                                           | Key properties                                                                                      |
-| --------------------------- | --------------- | ----------------------------------------------- | --------------------------------------------------------------------------------------------------- |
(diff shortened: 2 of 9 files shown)


bonevbs (commented):
More ov a high-level comment but in some sense this could be considered truncation logic. So perhaps it would be good to move this to where we have the sht truncation logic? What do you think @azrael417

azrael417:
We fix this in the big grid refactor MR. There we have a nice place for this.

bonevbs (approved):
LGTM

PR #251, opened 2026-08-25
 Grid descriptors: replace (nlat, nlon, grid_string) with a GridS2 object

azrael417:
## Problem

Every layer took a resolution and a grid name as separate arguments, and then reconstructed the grid's geometry itself. The same `if grid == "legendre-gauss" / "lobatto" / "equiangular"` chain, followed by a hand-rolled `flip(arccos(...))`, appeared twelve times. The `theta_cutoff` default `π/(nlat−1)` appeared in six independent places — an expression that is only correct for equiangular grids.

Two consequences:

1. Adding a grid type means editing every consumer. A reduced Gaussian grid has no single `nlon`, so nothing that assumes a rectangular `(nlat, nlon)` can carry it.
2. The duplicated assumptions were already wrong. On a lobatto grid at `nlat=129`, 6 output latitudes saw only the ring they sit on; on trapezoidal, 30 did. The stencil silently degenerated at the poles.

A descriptor makes the grid one argument that knows its own nodes, weights, spectral bounds, and decomposition.

## The design — this is the part worth reviewing

`torch_harmonics/grid.py` (new, ~1430 lines) is the whole proposal. Everything else follows mechanically. Three decisions are load-bearing.

**1. Identity is a tuple of scalars, not tensors.** A descriptor is a frozen dataclass whose `key` is `(grid_type,) + its own fields`, backing both `__hash__` and `__eq__`. Nodes and weights are properties, never fields — a descriptor holding tensors would fall back to identity hashing and silently defeat every `lru_cache` keyed on a grid. `key` is derived from the dataclass fields rather than listing `nlat`/`nlon`, so a family parameterized by a HEALPix `nside` or a refinement level gets correct identity without registering anything.

**2. The descriptor states facts; `truncation.py` makes decisions.** `max_exact_degree`, `max_azimuthal_order`, `max_node_spacing` report what a grid can represent and never warn. `truncate_sht` and `truncate_support` apply overrides, pick defaults, and announce changed ones. The grid deliberately has no `theta_cutoff` at all: a grid has a *spacing*, which is a fact; turning it into a *cutoff* is policy.

**3. Three levels, each exactly what some algorithm may assume.**

| level | adds | which buys |
| --- | --- | --- |
| `PointSetS2` | points and weights | integration |
| `GridS2` | isolatitude rings, equispaced in longitude | an FFT in longitude (hence a fast SHT), a neighbourhood search bounded by a search over rings, a contiguous polar decomposition |
| `RegularGridS2` | every ring the same length | a dense `(nlat, nlon)` layout, the compiled kernels, a 2D decomposition |

Every subclass adds a constraint, so every `RegularGridS2` is a `GridS2` is a `PointSetS2`. A routine states which contract it needs by calling `require_point_set`, `require_grid` or `require_regular_grid`; today 44 call sites demand regular and 4 demand only points; none yet need rings-but-not-regular. Each guard relaxes independently as a backend gains support — they are meant to come down one at a time, not all at once.

This is what makes a ragged grid (reduced Gaussian, HEALPix) or an unstructured one (an ICON icosahedral mesh) an additive change rather than a second API break.

`GridShardS2` is deliberately **not** a `GridS2` — a latitude band does not cover S², its weights do not sum to 2, and it defines no global quantity at all. That is a correctness argument, not tidiness: a support radius derived from a shard's own spacing would differ between ranks, and ranks disagreeing about an operator's support is a bug, not an inefficiency.

## Files to review, in order

- **`torch_harmonics/grid.py`** — the design itself; everything else is a consequence of it. Start here.
- **`torch_harmonics/truncation.py`** — the fact/policy boundary, and both truncation routines.
- **`torch_harmonics/integration.py`** — small, and the one consumer that has actually been taken ragged. `QuadratureS2` accepts any `PointSetS2` and derives its reduction axes from `grid.shape`, so it works unchanged on a flat `(npoints,)` field.
- **`torch_harmonics/distributed/distributed_convolution.py`** — the sharding conversion where it is subtlest: psi is built globally and then split, and the forward deliberately keeps `nlat_out` whole while the transpose keeps `nlat_in` whole.
- **`tests/test_grid_assumptions.py`** — the contract the descriptor has to satisfy. Two stand-in grids carry most of the weight: `_RaggedGrid` (rings of unequal length) and `_FibonacciPointSet` (no rings at all). They are defined in the tests rather than shipped, and they exist so that the day a ring assumption leaks into the base class, something fails.
- `torch_harmonics/partition.py` — trivial; exists to break an import cycle and give the descriptors and the distributed primitives one `compute_split_shapes`.

Skim-only: the SHT, attention and resample layers are the same substitution repeated, and the notebook changes are call-site updates.

## Breaking changes

**Layers take a descriptor.**

```python
RealSHT(nlat, nlon, lmax=..., grid="legendre-gauss")
  -> RealSHT(as_grid("legendre-gauss", nlat=nlat, nlon=nlon), lmax=...)

SpectralConvS2(in_shape, out_shape, in_channels, out_channels, grid_in=..., grid_out=...)
  -> SpectralConvS2(grid_in, grid_out, in_channels, out_channels)
```

All 14 public layers now lead with `grid` / `grid_in, grid_out`. Channels-first could never be uniform — `QuadratureS2`, the four SHTs and `ResampleS2` have no channels — so grid-first is the only ordering that yields one rule. The break is loud: an int landing in `grid_in` dies in `require_grid` naming the parameter.

**`as_grid` takes keywords**, not a positional shape: `as_grid("equiangular", nlat=128, nlon=256)`. Parameters are validated against the grid type, so one that does not apply raises rather than being ignored.

**`lats` changed meaning — the one that does not announce itself.** `GridS2.lats` returned *colatitude* despite the name. It is now `colats`, and `lats` returns geographic latitude `π/2 − θ`. Code reading `grid.lats` keeps running and gets an angle reflected about the equator, so it must be updated rather than left. Three call sites in this repo were already hand-writing `π/2 - grid.lats`. `quadrature.precompute_latitudes` is unchanged and still returns colatitudes.

**The two weight tensors split.** `GridS2.colat_weights` is the latitudinal factor, shape `(nrings,)`, summing to 2. `PointSetS2.quad_weights` is the per-point solid angle, shape `(npoints,)`, summing to 4π. They differ by exactly the longitudinal factor; the shapes differ too, so substituting one for the other raises rather than scaling an integral by 2π silently.

**`"equiangular-trapezoidal"` is renamed `"trapezoidal"`** (class `TrapezoidalGrid`). The old name described nodes it does not have — equispaced in `cos θ`, not in `θ`, up to 19° from the equiangular grid's. The old string is no longer accepted.

**The example losses and metrics take a descriptor**: `L2LossS2(nlat=n, nlon=2*n, grid="equiangular")` becomes `L2LossS2(as_grid("equiangular", nlat=n, nlon=2*n))`.

`GaussianRandomFieldS2` no longer assumes `nlon = 2*nlat`.

## Numerical behavior

**Inert where it should be.** Every quadrature and Legendre buffer and all transform outputs are bit-identical; equiangular psi is unchanged in both sparsity and values. The distributed quadrature weights are bit-identical to the previous construction on every rank of 1×1, 2×2, 4×1 and 1×4.

**Deliberately changed — `theta_cutoff` defaults.** Two separate changes, both affecting only callers who do *not* pass `theta_cutoff` explicitly:

1. The default comes from the grid's node distribution rather than `π/(nlat−1)`. Equiangular moves by ~1e-15; lobatto widens ~21%, legendre-gauss narrows ~2.3%, trapezoidal widens by `2√(nlat−1)/π` — ~5× at `nlat=65`, ~17× at 721.
2. The default is one *node* spacing rather than one *latitudinal* spacing — the grid's coarsest neighbour distance, in whichever direction it lies. Equiangular at `nlon = 2·nlat` is unaffected; legendre-gauss at the same resolution widens 0.75%; `nlon < 2·nlat` widens by 2× or more. Across the six grid/resolution combinations I measured, including the largest in the suite (721x1440 -> 360x720), the support band stays 3 rings, so sparsity patterns and halo radii are unchanged and only psi *values* shift.

`UserWarning` on the affected grids, naming which of the two reasons applies. Pass `theta_cutoff` explicitly for the old behaviour — an explicit value is returned verbatim and is unaffected by either change.

## Testing

```
pytest tests/                              # 2293 passed, 24 skipped, 843 subtests
torchrun ... tests/test_distributed_*.py   # H, W ∈ {1,2,4}
```

Three new test files, and two stand-in grids that exist purely to hold the abstraction honest: `_RaggedGrid` proves the ring-ragged path, `_FibonacciPointSet` proves the ringless one. The latter found a real bug on its first run — `truncate_support` was guarding at `require_grid` when it needs only `require_point_set`.


Changelog.md
bonevbs, line 1: the changelog is much too verbose for my taste. Please stick to the prior style of short, descriptive bullet points.

bonevbs: also this should be added for a major bump in the torch-harmonics version

torch_harmonics/grid.py
```diff
@@ -0,0 +1,733 @@
...
+
+    grid_type: ClassVar[str] = "lobatto"
+
+    @property
+    def max_exact_degree(self) -> int:
+        r"""Gauss--Lobatto is exact to degree :math:`2N_\theta - 3`, giving :math:`N_\theta - 1`."""
+        return self.nlat - 1
+
+
+@dataclass(frozen=True, eq=False)
```
bonevbs, line 777: I wouldn't use an extra type for this. It is the same grid but a different quadrature rule. we can perhaps just add the special quadrature as an option to the regular equiangular grid

bonevbs (commented):
Leaving some comments. More tomorrow.

torch_harmonics/grid.py
bonevbs: I mean the equiangular trapezoidal grid class one below

azrael417: it is a different grid, no? the lat coordinates are slightly differently spaced

azrael417: since trapezoidal is equispaced in cos theta and not equispaced in theta.

azrael417: I renamed it to trapezoidal

azrael417 (commented):
please let me know what you think

Changelog.md
azrael417: fixed

PR #253, opened 2026-09-02
Potential Security Vulnerabilities Fixes

azrael417:
Three findings from a security scan. Two are hardening with no behaviour change; the third fixes a genuine memory-safety bug.
  
  weights_only=True for example checkpoints — examples/{depth,segmentation,shallow_water_equations}/train.py. Each script saves model.state_dict(), so the stricter loader accepts existing
  checkpoints unchanged and --resume still works. Note this is already the default on torch>=2.6, which pyproject.toml requires, so the flag is explicit rather than newly protective.
  
  Pin NATTEN in Dockerfile.examples to 39375ea1d2f3b1bb284315fa5f79f6ecbebb6382 (v0.21.7) instead of cloning the default branch. A SHA rather than the tag alone, since a tag can be moved;
  git verifies the object hash on checkout. That SHA is what upstream HEAD resolved to when the pin was added, so the image builds identically today. Also dropped make WITH_CUDA=1 —
  NATTEN's Makefile never reads that variable; CUDA_ARCH is what selects the CUDA build.
  
  Validate index tensors in preprocess_psi — disco_helpers.{cpp,h}. Koff[ker_h[i]]++ indexed a K-sized heap buffer with unchecked values, and nnz was taken from val.size(0) while the
  index tensors' lengths were never checked. It now validates dtype, K, Ho > 0, matching 1-D lengths, and ker_idx ∈ [0,K), row_idx ∈ [0,Ho) — mirroring what the sibling pack_psi_dense
  already did. The Ho*K overflow path also called exit(EXIT_FAILURE), terminating the host interpreter from inside a Python extension; it now raises via TORCH_CHECK.
  
  Testing: pytest tests/test_convolution.py tests/test_distributed_convolution.py — every DiscreteContinuousConvS2 construction goes through preprocess_psi, so a too-strict check surfaces
  there. Before the fix, crafted inputs produced SIGSEGV, SIGBUS, and malloc(): invalid next size (heap metadata corruption); after, each raises a RuntimeError, and valid plus empty
  (nnz=0) inputs are unaffected.
  
  On severity: the preprocess_psi issue crosses no trust boundary. It is only called from __init__ with indices computed by _precompute_convolution_tensor_s2, never with deserialized or
  user-supplied data, and anyone able to call it with crafted tensors already executes Python in the process. Worth fixing as robustness — a caller mistake should raise rather than
  corrupt the heap — but not an exploitable vulnerability.

### DIFF
--- a/torch_harmonics/disco/optimized/disco_helpers.cpp
+++ b/torch_harmonics/disco/optimized/disco_helpers.cpp
@@ -41,6 +41,11 @@ torch::Tensor preprocess_psi(const int64_t K, const int64_t Ho, torch::Tensor ke
     CHECK_CONTIGUOUS_TENSOR(col_idx);
     CHECK_CONTIGUOUS_TENSOR(val);
 
+    TORCH_CHECK(ker_idx.dtype() == torch::kInt64, "ker_idx must be int64");
+    TORCH_CHECK(row_idx.dtype() == torch::kInt64, "row_idx must be int64");
+    TORCH_CHECK(col_idx.dtype() == torch::kInt64, "col_idx must be int64");
+    TORCH_CHECK(K > 0 && Ho > 0, "K and Ho must be positive");
+
     // get the input device and make sure all tensors are on the same device
     auto device = ker_idx.device();
     TORCH_INTERNAL_ASSERT(device.type() == row_idx.device().type() && (device.type() == col_idx.device().type())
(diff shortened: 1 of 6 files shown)


torch_harmonics/disco/optimized/disco_helpers.cpp
```diff
@@ -41,6 +41,11 @@ torch::Tensor preprocess_psi(const int64_t K, const int64_t Ho, torch::Tensor ke
     CHECK_CONTIGUOUS_TENSOR(col_idx);
     CHECK_CONTIGUOUS_TENSOR(val);
 
+    TORCH_CHECK(ker_idx.dtype() == torch::kInt64, "ker_idx must be int64");
```
bonevbs, line 44: we never expect to get kint32 here I presume?

bonevbs (approved):
looks mostly good, just one comment - please have a look before merging

torch_harmonics/disco/optimized/disco_helpers.cpp
azrael417: we should not

azrael417: otherwise we could overflow for huge tensors

azrael417: yes, internally the kernel assumed int64, so it would have silently read garbage if the data were in int32. so this fix is super important actually.

PR #255, opened 2026-09-03
Fix the docstring examples that never ran

azrael417:
No module in the package has ever had its docstring examples executed: pytest collects only tests/, and neither pyproject.toml nor the CI workflow passes --doctest-modules. The examples are still rendered into the API docs by autodoc, so they read as verified output while being unchecked. Running them turns up 7 failures with CUDA and 21 without.

The SHT examples documented the pre-v0.9.0 truncation. On a 128x256 equiangular grid lmax is (nlat+1)//2 = 64 and triangular truncation gives mmax = 64, so RealSHT returns [1, 64, 64] rather than [1, 128, 129]. The two inverse examples did not merely print the wrong shape, they raised: they built coefficients at the documented shape and the layer's own check rejected them. Those two now size their input from isht.lmax/isht.mmax, since a hardcoded shape is what drifted in the first place.

QuadratureS2 documented the exact value of 4*pi where the code returns the float32 result, its weights buffer being float32. Rounded to five places, which is stable in both precisions rather than pinning float32 digits.

The remaining failures were the examples pinned to CUDA, which cannot run in CPU-only CI and cannot be copy-pasted by a reader without a GPU. They demonstrate shapes, not device placement, so the .cuda() and device="cuda" are dropped.

Docstring text only; no behaviour changes. What the docstrings got wrong is covered correctly by existing tests in tests/ in every case.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,6 +36,16 @@ jobs:
         python3 -m pytest -ra --cov-report term --cov-report json --cov-config=.coveragerc --cov=torch_harmonics \
           --ignore-glob="**/test_distributed_*" ./tests/
 
+    # Execute the ">>>" examples in the package docstrings. Sphinx renders them
+    # into the API docs, where they read as verified output, so an unchecked one
+    # is documentation that lies: before this step existed every example in the
+    # package was inert, and several had been wrong for releases. Kept separate
+    # from the suite above so that a broken example is its own red mark and does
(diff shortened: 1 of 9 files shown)


albertocarpentieri (commented):
why did you remove the use_mlp argument?

azrael417:
use_mlp is a dead argument on those models

albertocarpentieri (approved):

PR #257, opened 2026-09-09
DISCO CUDA: faster kpacked forward, transposed-ownership backward, compact psi

azrael417:
Optimizes both DISCO CUDA paths around the same observation: psi's sparsity per output row is a contiguous longitude arc, so the index arithmetic that dominated both kernels is mostly recomputing a shift that is already known.

Forward (kpacked, sm_90a + sm_100a)
-----------------------------------
The kernel was instruction-issue bound, not memory bound: 1.68G warp instructions against 2.37G issue slots (70.9%, matching Compute (SM) Throughput and Issued Warp Per Scheduler exactly), with only 6% of instructions being global loads. Staging 8 elements cost ~153 instructions per warp per chunk, nearly all of it address computation.

  * Windowed A-tile gather. Because the arc is contiguous, one aligned 32-bit window covers the whole staging chunk, and the per-element addresses collapse into shifts of it -- __funnelshift_r for pscale 1, __byte_perm for pscale 2 (gather_window_rt in disco_cuda.cuh, shared by both arch kernels). Removes ~35% of instructions. Gated to pscale <= 2; larger pscale keeps the general path.
  * Vectorized pack_idx load (longlong2).
  * Blocked-CSR psi: row offsets over ho with K-packed value blocks, so NBR_PAD no longer pads every row to the global maximum.

              H100                      GB200
  pscale 1    3.22 -> 2.65 ms  1.22x    2.38  -> 1.855 ms  1.28x
  pscale 2   11.04 -> 10.61 ms 1.04x    6.94  -> 6.21  ms  1.12x

pack_val footprint: 33 MB -> 1.9 MB at half degree, ~410 MB -> 24 MB at 1080x2160.

Backward
--------
disco_bwd_d gave each thread a slice of the input row in registers and accumulated into a shared output row, making every nonzero a shared read-modify-write plus a __syncthreads to stop consecutive nonzeros racing. MIO throttle and barrier stalls together were 62% of blocking stalls, while shared wavefronts already equalled the ideal count -- no waste to reclaim, only volume to remove.

disco_bwd_swap_d transposes the ownership. Shared holds the input row, written once and read-only after; the accumulator lives in registers. The inner loop becomes LDS+FFMA instead of LDS+FADD+STS, and the only barrier left is the one after the shared fill.

                          H100                     GB200
  prod_decoder (ps 1)  121.6 -> 59.19 ms 2.05x   83.21 -> 39.94 ms 2.08x
  prod_encoder (ps 2)   40.20 -> 21.97 ms 1.83x   27.38 -> 14.20 ms 1.93x

Two details that carry most of the gain and are easy to lose:

  * The pscale bank select is the OUTER loop with the whole element loop inside it. Written the other way round, with a one-instruction body, ptxas predicates instead of branching -- correct for a body that small, but it then issues PSCALE adds per element. That cost +28% instructions and turned a memory-bound kernel into an issue-bound one: 42.25 ms against 24.45 ms for the current nesting, from loop order alone.
  * Gated to pscale <= 2. The accumulator costs PSCALE*ELXTH registers; at pscale 3 (36 registers at Wi=720) occupancy falls from 36.6% to 22.4% and the kernel loses despite executing 10% fewer instructions. pscale 3 with a smaller ELXTH (18 registers) still wins slightly, so the real predictor is register count -- pscale is just what can be gated cheaply.

Backward launch geometry
------------------------
The dispatch selected the block shape from Wo, inherited from the forward where the output row does live in shared. The backward's only geometric requirement is NTH*ELXTH >= Wi. Whenever Wo > 64*ELXTH_MAX >= Wi the NTH=128 branch started ELXTH at 17, overshooting by 3x, and multiplied by pscale that overran the 48 KB static shared limit: any shape with Wo > 2048 and pscale >= 3 failed to launch at all (e.g. 1080x2160 -> 360x720, 52224 B).

Keyed on Wi instead. For Wo <= 64*ELXTH_MAX this is provably a no-op, since Wi <= Wo puts both on the same branch with identical growth; it only changes shapes that were previously broken or wasteful. An explicit guard now reports the shape and the limit rather than surfacing as a bare invalid argument.

Also updates benchmarks/reference_results.csv and adds benchmarks/reference_results_updated.csv.

Correctness: full test suite green on H100 and GB200. Kernel-level checks ran against the unmodified bodies at fp32 across pscale 1, 2 and 3 (hdeg_self, hdeg_down, qdeg_down, 1deg_down3, hdeg_down3, prod_encoder, prod_decoder); every one sat at the run-to-run noise floor of the atomicAdd ordering, ~4e-7.


rietmann-nv (approved):
Looks faster to me

PR #260, opened 2026-09-10
Fix inductor codegen of the complex-valued SHT and spectral convolution paths

azrael417:
## Problem

  `torch.compile` of any model containing `SpectralConvS2` (or the SHT layers directly)
  fails on CUDA during inductor codegen:

  File "torch/_inductor/codegen/triton_utils.py", line 48, in signature_of
    typ = _type_of(arg.dtype)
  File "torch/_inductor/utils.py", line 507, in _type_of
    return key if isinstance(key, str) else f"*{tys[dtype_str]}"
  torch._inductor.exc.InductorError: KeyError: 'complex64'

  Triton has no complex type, so inductor cannot generate a kernel that reads or writes a
  complex buffer. Complex ops are expected to fall back to eager extern kernels, but several
  pointwise ops in these layers were lowered into Triton kernels instead, and codegen died
  when building the kernel signature.

  The transforms also had a second, latent exposure: `aten.complex` is a fallback kernel, and
  inductor's meta for it predicts a contiguous output. Given non-contiguous operands, eager's
  TensorIterator allocates a permuted output instead, and the emitted `assert_size_stride`
  fires. The scalar transforms had already been fixed for this; the vector ones had not, and
  the distributed vector transforms applied their `.contiguous()` after `torch.stack`, which is
  too late — the assert fires on `aten.complex`'s own output.

  ## Approach

  Keep the complex dtype off every pointwise op. Scaling, stacking, contiguity and reshapes now
  run on real tensors (directly, or via `view_as_real`/`view_as_complex`, which are pure views),
  leaving `aten.complex` and the FFTs — both fallbacks — as the only complex-typed nodes.

  `view_as_real` is specifically the one consumer for which inductor's complex-tensor check
  makes an exception, so this is the sanctioned way to hand complex data to real codegen. Note
  that none of it reintroduces the in-place writes that motivated removing `view_as_complex`
  from these layers in `6f7706a`: every use here is out-of-place.

  ## Changes

  **`sht.py`, `distributed/distributed_sht.py`**

  - Fold the `2*pi` longitudinal scale factor of the forward-normalized FFT into the precomputed
    quadrature weights, instead of scaling the complex FFT output on every call. It is a constant
    prefactor of a linear transform, so this is exact, and it removes a complex multiply from the
    hot path of all four forward transforms.
  - The vector transforms now stack the spheroidal/toroidal components in real space and assemble
    the result with a single `torch.complex` over the (contiguous) stack outputs. This drops a
    complex `cat` and fixes the `assert_size_stride` exposure in one step. The now-dead
    `.contiguous()` after the stack in the distributed versions is gone.

  **`utils.py`**

  - `_EnsureContiguous` performs its copy on the real view for complex tensors. The backward is
    load-bearing — an upstream layer can hand back a non-contiguous gradient that the CPU/MKL FFT
    backward rejects (`DftiCommitDescriptor`) — so it is preserved, just no longer complex-typed.
  ## Changes

  **`sht.py`, `distributed/distributed_sht.py`**

  - Fold the `2*pi` longitudinal scale factor of the forward-normalized FFT into the precomputed
    quadrature weights, instead of scaling the complex FFT output on every call. It is a constant
    prefactor of a linear transform, so this is exact, and it removes a complex multiply from the
    hot path of all four forward transforms.
  - The vector transforms now stack the spheroidal/toroidal components in real space and assemble
    the result with a single `torch.complex` over the (contiguous) stack outputs. This drops a
    complex `cat` and fixes the `assert_size_stride` exposure in one step. The now-dead
    `.contiguous()` after the stack in the distributed versions is gone.

  **`utils.py`**

  - `_EnsureContiguous` performs its copy on the real view for complex tensors. The backward is
    load-bearing — an upstream layer can hand back a non-contiguous gradient that the CPU/MKL FFT
    backward rejects (`DftiCommitDescriptor`) — so it is preserved, just no longer complex-typed.

  **`distributed/primitives.py`**

  - `flatten_and_pad_leading_dims` / `unpad_and_unflatten_leading_dims` do their layout work on the
    real view when handed complex input. These are pure leading-dim helpers, so the appended re/im
    axis is simply carried along. Covers the `reshape`, the zero-padding `cat` and the `contiguous`.

  **`spectral_convolution.py`, `distributed/distributed_spectral_convolution.py`**

**`spectral_convolution.py`, `distributed/distributed_spectral_convolution.py`**

- Dropped `.contiguous()` after `self.sht(x)`; the SHT output is now contiguous by construction.
- The spectral bias is applied on the real view. `integral` is computed from the real-valued
  input, so it is a real tensor broadcasting over the re/im axis and the rewrite is exact.
- The post-contraction reshape merges the group axes on the real view. Unlike the `sht` case the
  `.contiguous()` is kept: the einsum output can be non-contiguous, so this reshape may copy —
  the point is only that the copy is real-dtype.

## Tests

- `tests/test_sht.py`: `test_compile` in the scalar and vector serial classes — round trip under
  `fullgraph=True`, forward and backward against eager, three grid/parity cases each. The round
  trip is compiled as one function so the coefficients are an intermediate buffer rather than a
  graph output, which is the case inductor has to codegen.
- `tests/test_spectral_convolution.py`: `test_compile`, four cases over bias on/off and
  grouped/ungrouped. `fullgraph` is not requested here since `_contract_lwise` carries its own
  `@torch.compile`; the claim is that codegen succeeds, not that the layer forms a single graph.

Both classes are parameterized over CPU and CUDA. The `KeyError` is Triton-specific, so the CUDA
runs are what cover it; CPU exercises the C++ backend, where the exposure is the stride assert.

## Verification

`tests/test_spectral_convolution.py` passes on CPU and CUDA (54 tests). The SHT and distributed
tests still need to be run.

The distributed changes have no compile coverage by design and rest on the existing correctness
tests — including the `B*C = 1` / `B*C = 2` rows, which are what exercise the zero-padding branch
on complex input under a real process grid.

## Notes for reviewers

- The `2*pi` fold is the one change that could silently alter numerics if mis-sited. The accuracy
  tests (`test_known_function`, `test_parseval`, the vector consistency tests) are the guard.
- The SHT `weights` buffers are all `persistent=False`, so folding the constant into them has no
  checkpoint-compatibility implication.
- Downstream models carrying their own copy of `spectral_convolution.py` (e.g. makani) need the
  same three edits; the change here does not reach them.
  runs are what cover it; CPU exercises the C++ backend, where the exposure is the stride assert.

  ## Verification

  `tests/test_spectral_convolution.py` passes on CPU and CUDA (54 tests). The SHT and distributed
  tests still need to be run.

  The distributed changes have no compile coverage by design and rest on the existing correctness
  tests — including the `B*C = 1` / `B*C = 2` rows, which are what exercise the zero-padding branch
  on complex input under a real process grid.

  ## Notes for reviewers

  - The `2*pi` fold is the one change that could silently alter numerics if mis-sited. The accuracy
    tests (`test_known_function`, `test_parseval`, the vector consistency tests) are the guard.
  - The SHT `weights` buffers are all `persistent=False`, so folding the constant into them has no
    checkpoint-compatibility implication.
  - Downstream models carrying their own copy of `spectral_convolution.py` (e.g. makani) need the

### DIFF
--- a/torch_harmonics/distributed/primitives.py
+++ b/torch_harmonics/distributed/primitives.py
@@ -201,16 +201,25 @@ def flatten_and_pad_leading_dims(tensor: torch.Tensor, min_leading_size: int, nu
     lead_size : int
         The true (pre-pad) flattened leading size, used to slice off the padding.
     """
+    # complex inputs (the inverse SHTs call this on their spectral input) are processed as
+    # their real view, see unpad_and_unflatten_leading_dims for the rationale
+    is_complex = tensor.is_complex()
+    if is_complex:
+        tensor = torch.view_as_real(tensor)
+        num_trailing_dims += 1
+
     lead_shape = tensor.shape[:-num_trailing_dims]
     tensor = tensor.reshape(-1, *tensor.shape[-num_trailing_dims:])
     lead_size = tensor.shape[0]
 
     if lead_size < min_leading_size:
-        # new_zeros preserves dtype (incl. complex) and device
+        # new_zeros preserves dtype and device
         zeros = tensor.new_zeros((min_leading_size - lead_size, *tensor.shape[1:]))
         tensor = torch.cat([tensor, zeros], dim=0)
 
-    return tensor.contiguous(), lead_shape, lead_size
+    tensor = tensor.contiguous()
+
+    return (torch.view_as_complex(tensor) if is_complex else tensor), lead_shape, lead_size
 
 
 def unpad_and_unflatten_leading_dims(tensor: torch.Tensor, lead_shape, lead_size: int, num_trailing_dims: int = 2) -> torch.Tensor:
@@ -219,9 +228,24 @@ def unpad_and_unflatten_leading_dims(tensor: torch.Tensor, lead_shape, lead_size
     The trailing ``num_trailing_dims`` dims are taken from ``tensor`` as-is, so this is
     valid even when the transform changed them (e.g. ``nlat, nlon`` -> ``lmax, mmax``).
     ``num_trailing_dims`` must match the value passed to the flatten call.
+
+    Complex inputs (the forward SHTs call this on their spectral output) are processed as
+    their real view: this is pure layout code touching only the leading dims, and a copy
+    over a complex buffer cannot be codegen'd by inductor -- triton has no complex type,
+    so it fails with ``KeyError: 'complex64'``. ``view_as_real`` appends the re/im pair as
+    a trailing dim, which is simply carried along; the ``contiguous()`` right before the
+    ``view_as_complex`` supplies the unit last-dim stride the latter requires.
     """
+
+    is_complex = tensor.is_complex()
+    if is_complex:
+        tensor = torch.view_as_real(tensor)
+        num_trailing_dims += 1
+
     tensor = tensor.narrow(0, 0, lead_size)
-    return tensor.reshape(*lead_shape, *tensor.shape[-num_trailing_dims:]).contiguous()
+    out = tensor.reshape(*lead_shape, *tensor.shape[-num_trailing_dims:]).contiguous()
+
+    return torch.view_as_complex(out) if is_complex else out
 
 
 def _transpose(tensor, dim0, dim1, dim1_split_sizes, group=None, async_op=False, verify_shapes=None):
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,6 +6,7 @@
 
 * Added `torch.compile(fullgraph=True)` support to the serial and distributed layers, which previously broke on assertion helpers and on control flow that branched on tensor values.
 * The forward SHTs now fold the `2*pi` longitudinal scale factor into their precomputed quadrature weights instead of scaling the FFT output on every call, which removes a pointwise multiply over a complex tensor from the hot path.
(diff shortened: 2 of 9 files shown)


torch_harmonics/distributed/primitives.py
```diff
@@ -219,9 +228,24 @@ def unpad_and_unflatten_leading_dims(tensor: torch.Tensor, lead_shape, lead_size
...
+    is_complex = tensor.is_complex()
+    if is_complex:
+        tensor = torch.view_as_real(tensor)
+        num_trailing_dims += 1
+
     tensor = tensor.narrow(0, 0, lead_size)
-    return tensor.reshape(*lead_shape, *tensor.shape[-num_trailing_dims:]).contiguous()
+    out = tensor.reshape(*lead_shape, *tensor.shape[-num_trailing_dims:]).contiguous()
+
+    return torch.view_as_complex(out) if is_complex else out
```
bonevbs, line 248: why is this needed?

bonevbs (commented):
looks mostly good. only the optional cast is puzzling me. Can you elaborate?

torch_harmonics/distributed/primitives.py
azrael417: to avoid doing anything in complex64 datatype. so first make config then return as view.

bonevbs (approved):
LGTM

PR #261, opened 2026-09-11
# Distributed SHT: distributed matmul over l/k, and per-rank Legendre construction

azrael417:
Two changes that together make the distributed SHT's memory scale with the process grid
  instead of with resolution alone, plus a precision fix in the reduction primitives.

  No API change. The serial transforms are untouched and bitwise unchanged.

  ## Motivation

  The precomputed Legendre coefficients are an `(mmax, lmax, nlat)` float64 table — under
  triangular truncation, `nlat³ · 8` bytes:

  | nlat | table |
  |---|---|
  | 512 | 1.1 GB |
  | 1024 | 8.6 GB |
  | 2048 | 69 GB |

  Two problems. The table was split only along `m`, so it was **replicated across the entire
  polar group**; and every rank *built the whole thing* before discarding the part it didn't
  own. At the resolutions we want, construction OOMs before the transform ever runs.

  ## Precision changes — please read this part

  **The distributed SHT no longer agrees with the serial transform to near-bitwise
  precision.** This is inherent to the new decomposition, not a regression to be fixed later,
  and it is the main thing to weigh when reviewing.

  Previously each rank gathered the full latitude axis and performed the quadrature sum
  locally, so the distributed result matched the serial one to a bit or two. Now latitudes
  stay distributed, each rank sums its own chunk, and the partial sums are combined by a
  reduce-scatter. Reassociating a float32 sum across ranks perturbs the result by a few ULP
  **of the summands**.

  Measured at H=2, float32:

  | worst \|a−b\| | element magnitude | relative | fp32 ULP |
  |---|---|---|---|
  | 1.9e-6 | 7.00 | 2.7e-7 | ~2.3 |
  | 1.4e-6 | 2.29 | 6.2e-7 | ~5.2 |

  The subtlety worth internalizing: **the error scales with the terms being summed, not with
  the result.** The worst offenders are near-cancellation outputs — an element of magnitude
  0.015 in a tensor whose scale is 2, produced by summing terms two orders larger. Its error
  is inherited from those terms, so its *relative* error is ~1e-4 however the sum is arranged.

  That is why the test tolerances moved to `atol=1e-5, rtol=1e-6` rather than simply raising
  `rtol`. A relative criterion cannot cover near-cancellation outputs without asserting
  something false about small results; the criterion that matches the numerics is absolute and
  scaled to the tensor's dynamic range, `eps_fp32 · max|x| ≈ 1e-6`, with ~8× headroom. The
  spread grows slowly with the number of partial sums, so this is calibrated at H=2 and worth
  re-measuring if we start testing at large polar counts.

  **What did *not* change:** the coefficients themselves. The restricted Legendre construction
  is bitwise identical to the full one — verified 168/168 across both routines, all norms,
  `inverse` and `csphase`. The precision shift lives entirely in the collective, not in the
  tables, which keeps the two effects separable when something looks wrong.

  **A precision fix in the opposite direction:** `_reduce` and `_reduce_scatter` unconditionally
  converted to fp32 when `use_fp32=True`, which **silently demoted fp64 inputs**. The flag exists
  to stop fp16/bf16 losing precision the local reduction kept; demoting fp64 is the opposite of
  its intent. Both now go through `_reduction_dtype`, which promotes only from an explicit
  low-precision list and never demotes. This also affects the DISCO and attention paths, where
  fp64 all-reduces were being capped at fp32. Not covered by the suite — every distributed test
  is fp32 — so it rests on inspection.

  ## Distributed matmul over l/k

  `distributed_sht.py`: latitudes (forward) and degrees (inverse) stay split across the polar
  group. Each rank contracts its own chunk and a reduce-scatter completes the sum while landing
  the output partitioned on the right axis.

  **All eight polar all-to-alls are gone, replaced by four reduce-scatters** — one collective per
  transform instead of two. Flops are unchanged in total. Channels are never split by the polar
  group anymore, so the leading-dim padding drops from `max(polar, azimuth)` to just `azimuth`.

  The re/im pair is stacked into a trailing axis before the collective and viewed back to complex
  after. This is required, not cosmetic: the reduce-scatter reduces in fp32, and `.to(float32)`
  on a complex tensor discards the imaginary part.

  Resident weights per rank go from `(mmax/P_a, lmax, nlat)` to `(mmax/P_a, lmax, nlat/P_p)` —
  fully partitioned across the grid rather than replicated over the polar group.

  Two costs worth noting: the reduce-scatter input still spans the full `l` extent, so the
  *activation* temporary is `P_p×` larger than the data it carries; and the contraction dimension
  shrinks to `nlat/P_p`, so the gemm gets skinnier as the polar group grows. Neither blocks this
  MR, but they are what would motivate a ring over `l` or a 3D process grid later.

  ## Per-rank Legendre construction

  `legendre.py`: `legpoly` now **streams the degree axis**. The recurrence reads only `l-1` and
  `l-2`, so the `(nmax, nmax, len(x))` table was pure convenience; the working set is now the
  result plus three `(mmax, len(x))` planes. This helps the serial path too — `RealSHT` at
  nlat=1024 was building an 8.6 GB intermediate to keep an 8.6 GB result.

  Both `legpoly` and `_precompute_dlegpoly` take keyword-only `mmin`/`lmin`, defaulting to 0.
  The three axes behave differently and the docstrings record why:

  * **latitudes** — no parameter. Evaluation points are independent, so restricting them is just
  stay distributed, each rank sums its own chunk, and the partial sums are combined by a
  reduce-scatter. Reassociating a float32 sum across ranks perturbs the result by a few ULP
  **of the summands**.

  Measured at H=2, float32:

  | worst \|a−b\| | element magnitude | relative | fp32 ULP |
  |---|---|---|---|
  | 1.9e-6 | 7.00 | 2.7e-7 | ~2.3 |
  | 1.4e-6 | 2.29 | 6.2e-7 | ~5.2 |

  The subtlety worth internalizing: **the error scales with the terms being summed, not with
  the result.** The worst offenders are near-cancellation outputs — an element of magnitude
  0.015 in a tensor whose scale is 2, produced by summing terms two orders larger. Its error
  is inherited from those terms, so its *relative* error is ~1e-4 however the sum is arranged.

  That is why the test tolerances moved to `atol=1e-5, rtol=1e-6` rather than simply raising
  `rtol`. A relative criterion cannot cover near-cancellation outputs without asserting
  something false about small results; the criterion that matches the numerics is absolute and
  scaled to the tensor's dynamic range, `eps_fp32 · max|x| ≈ 1e-6`, with ~8× headroom. The
  spread grows slowly with the number of partial sums, so this is calibrated at H=2 and worth
  re-measuring if we start testing at large polar counts.

  **What did *not* change:** the coefficients themselves. The restricted Legendre construction
  is bitwise identical to the full one — verified 168/168 across both routines, all norms,
  `inverse` and `csphase`. The precision shift lives entirely in the collective, not in the
  tables, which keeps the two effects separable when something looks wrong.

  **A precision fix in the opposite direction:** `_reduce` and `_reduce_scatter` unconditionally
  converted to fp32 when `use_fp32=True`, which **silently demoted fp64 inputs**. The flag exists
  to stop fp16/bf16 losing precision the local reduction kept; demoting fp64 is the opposite of
  its intent. Both now go through `_reduction_dtype`, which promotes only from an explicit
  low-precision list and never demotes. This also affects the DISCO and attention paths, where
  fp64 all-reduces were being capped at fp32. Not covered by the suite — every distributed test
  is fp32 — so it rests on inspection.

  ## Distributed matmul over l/k

  `distributed_sht.py`: latitudes (forward) and degrees (inverse) stay split across the polar
  group. Each rank contracts its own chunk and a reduce-scatter completes the sum while landing
  the output partitioned on the right axis.

  **All eight polar all-to-alls are gone, replaced by four reduce-scatters** — one collective per
  transform instead of two. Flops are unchanged in total. Channels are never split by the polar
  group anymore, so the leading-dim padding drops from `max(polar, azimuth)` to just `azimuth`.

  The re/im pair is stacked into a trailing axis before the collective and viewed back to complex

  Two notes on the draft. The precision section leads with the trade-off rather than burying it, since that's the part a reviewer has to accept rather than verify. And it separates the two directions deliberately — the distributed transform got less precise by design, while the fp64 reduction fix made the collectives more precise — because conflating them would make the tolerance changes look like they were papering over the bug.

                                                                                                                                                     Update available! Run: brew upgrade claude-code
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
❯
─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
  ⏸ manual mode on · ? for shortcuts · ← for
The four distributed constructors now request exactly their block. `split_tensor_along_dim` is
gone from the file — nothing is built and discarded.

## Tests

**Serial** (`test_sht.py`), all new and none requiring a process grid — the streaming and the
ranges are now covered by ordinary CI rather than only under multi-rank:

* `test_addition_theorem` — `Σ_m |Y_l^m|² = const`, pointwise in θ, to `lmax=48`, all three
  norms. No quadrature, no transform, no reference table; at each degree it constrains every
  order at once. This is the check that reaches arbitrary `lmax`, where the existing closed
  forms stop at degree 3.
* `test_discrete_orthogonality` — full Gram matrix per order across three grids.
* `test_dlegpoly_consistency` — `dpct[1] = m·P/sinθ` (built from a different recurrence, so
  agreement at 5e-15 is genuine corroboration) and `dpct[0]` against a central difference.
* `test_range_restriction` — 1176 restricted blocks, bitwise.

**Distributed** (`test_distributed_sht.py`): `test_legendre_blocks` has each rank compare its
own buffer to the cut-out of the serial construction — no collective, so a failure identifies
the rank and separates a construction bug from a transform bug. Offsets are recomputed in the
test from the shape lists rather than read off the transform, so a wrong offset can't mask
itself.

Also flipped `test_gradient_consistency` to `verbose=False` and removed 20 `if verbose:
print("Testing ...")` banners from `test_sht.py`. Those fired on success and duplicated the
pytest test id; the rest of the suite uses `verbose` for timings and failure diagnostics only.

## Verification

* `tests/test_distributed_sht.py`: 75 passed across 6 grid configurations.
* Legendre: 168/168 bitwise identical to main on the default range; 1656/1656 restricted
  blocks equal the corresponding slice exactly; 600/600 simulated rank blocks match the
  undistributed construction across 7 grid shapes including uneven splits.
* The new serial tests were exercised through a standalone harness, not pytest.

## Follow-ups, not in this MR

* Ring over `l` to bound the reduce-scatter's full-`l` temporary.
* 3D process grid (`P_polar = P_l × P_k`) if the skinnier gemm shows up in benchmarks.
* A CHANGELOG entry — this branch doesn't add one yet.
    passing a shorter `t`.
  * **orders** — `mmin`, because the sectoral seed couples them: reaching `P^m_m` means walking up
    from `m=0`, storing nothing until the range begins.
  * **degrees** — `lmin`, same structure via the three-term recurrence. Needed by the inverse
    transforms, which are split along `l`.

  `_precompute_dlegpoly` requests its own halo (orders `m±1`, degrees `l` and `l+1`), so callers
  don't need to know about it.

  The four distributed constructors now request exactly their block. `split_tensor_along_dim` is
  gone from the file — nothing is built and discarded.

  ## Tests

  **Serial** (`test_sht.py`), all new and none requiring a process grid — the streaming and the
  ranges are now covered by ordinary CI rather than only under multi-rank:

  * `test_addition_theorem` — `Σ_m |Y_l^m|² = const`, pointwise in θ, to `lmax=48`, all three
    norms. No quadrature, no transform, no reference table; at each degree it constrains every
    order at once. This is the check that reaches arbitrary `lmax`, where the existing closed
    forms stop at degree 3.
  * `test_discrete_orthogonality` — full Gram matrix per order across three grids.
  * `test_dlegpoly_consistency` — `dpct[1] = m·P/sinθ` (built from a different recurrence, so
    agreement at 5e-15 is genuine corroboration) and `dpct[0]` against a central difference.
  * `test_range_restriction` — 1176 restricted blocks, bitwise.

  **Distributed** (`test_distributed_sht.py`): `test_legendre_blocks` has each rank compare its
  own buffer to the cut-out of the serial construction — no collective, so a failure identifies
  the rank and separates a construction bug from a transform bug. Offsets are recomputed in the
  test from the shape lists rather than read off the transform, so a wrong offset can't mask
  itself.

  Also flipped `test_gradient_consistency` to `verbose=False` and removed 20 `if verbose:
  print("Testing ...")` banners from `test_sht.py`. Those fired on success and duplicated the
  pytest test id; the rest of the suite uses `verbose` for timings and failure diagnostics only.

  ## Verification

  * `tests/test_distributed_sht.py`: 75 passed across 6 grid configurations.
  * Legendre: 168/168 bitwise identical to main on the default range; 1656/1656 restricted
    blocks equal the corresponding slice exactly; 600/600 simulated rank blocks match the
    undistributed construction across 7 grid shapes including uneven splits.
  * The new serial tests were exercised through a standalone harness, not pytest.

  ## Follow-ups, not in this MR

  * Ring over `l` to bound the reduce-scatter's full-`l` temporary.
  * 3D process grid (`P_polar = P_l × P_k`) if the skinnier gemm shows up in benchmarks.


bonevbs (approved):
LGTM

PR #264, opened 2026-09-16
Add SHT truncation modes

mwyau:
# Summary

Add `trapezoidal` and `rhomboidal` SHT truncation modes.

# Motivation

I'm adding a PyTorch / `torch-harmonics` backend to the [`spharmgrid`](https://github.com/mwyau/spharmgrid) atmospheric science regridding package, where different SHT truncation modes are supported.

#### References
* NCL/Spherepack supports both [triangular](https://www.ncl.ucar.edu/Document/Functions/Built-in/tri_trunc.shtml) and [rhomboidal truncation](https://www.ncl.ucar.edu/Document/Functions/Built-in/rhomb_trunc.shtml).
* [CAM6 Scientific Guide](https://acomstaff.acom.ucar.edu/tilmes/CAM_docs/doc/build/html/cam6_scientific_guide/chapter3.html#spectral-algorithm-overview) 

image

# Tests

* Default is triangular truncation.
* Benchmark done locally.
* Added truncation tests and parameterized tests in SHT tests.


azrael417 (commented):
Hello mwyau. 

Thanks for the PR, and for the clear motivation — independent `lmax`/`mmax` is a reasonable thing to want, and matching an existing spectral layout from NCL/Spherepack is a good reason to want it. Sorry for the slow turnaround.
I looked at the MR but also threw claude at it because I wanted to verify that it not accidentally breaks the distributed SHT. It should be fine. Here is a list of the findings:

My main worry was whether a non-triangular block would upset the distributed SHT, since that code was reworked fairly recently (#261). Good news: **it doesn't.** I checked it numerically rather than just by reading, on a 2x2 process grid, comparing the distributed transform against the serial one (`nlat=32, nlon=64`, Legendre-Gauss, scalar and vector, forward and inverse):

| lmax | mmax | fwd max err | inv max err |
|---|---|---|---|
| 24 | 8 | 4.5e-8 | 2.4e-7 |
| 32 | 8 | 4.5e-8 | 3.0e-7 |
| 32 | 16 | 6.0e-8 | 4.8e-7 |
| 16 | 16 (control) | 6.0e-8 | 2.4e-7 |

Rectangular agrees with the serial transform just as well as the triangular control, i.e. at the ULP level the reduce-scatter already costs us. That lines up with the structure of the code: `l` is split over the polar group and `m` over the azimuth group entirely independently, and the Legendre builders already take `mmax`/`lmax` as independent arguments with separate `mmin`/`lmin` windows. `lmax > nlat` was supported before (the "upsampling" rows in `test_distributed_sht.py` use `lmax=64` at `nlat=32`), so a non-square block isn't new territory for the splitting logic.

So the feature itself is sound. Two things I'd like to change before merging, plus some small stuff.

### 1. Could we make the rectangular mode explicit?

Right now the behaviour is inferred from *which arguments are present*, which produces a rule that's hard to explain:

```python
truncate_sht(16, 16, lmax=12)          # -> (9, 9)   lmax discarded
truncate_sht(16, 16, lmax=12, mmax=9)  # -> (12, 9)  lmax honoured
```

Passing an `mmax` that merely equals the default changes `lmax`. It also silently changes behaviour for anyone already calling with both limits and `lmax > mmax`: they used to get `(mmax, mmax)` and now get a different spectral shape. In the distributed case that's a bit more painful than it sounds, because `lmax` is split across the polar group, so `lmax_local` changes and existing `DistributedSpectralConv` weights/checkpoints no longer line up.

Would you be up for making it opt-in instead, something like `truncation="triangular"` (default) vs `"rectangular"`? Then triangular stays the fallback everywhere, no existing call changes behaviour, the `lmax`-alone asymmetry goes away, and the **Breaking** note in the Changelog can go with it. Happy to discuss the exact spelling — `"rhomboidal"` might map better onto the NCL naming you referenced.

### 2. A rectangular row in the distributed tests

Every row in `tests/test_distributed_sht.py` currently pins `mmax=lmax`, so nothing asymmetric is covered there. Since the serial tests you added are the natural place to start, could you add one rectangular case to the distributed parameterization too? Pick an `mmax` that divides the azimuth group size evenly and it slots straight in.

One thing to keep in mind if you test locally: the gloo backend can't do uneven all-to-all at all, so any `mmax` not divisible by the azimuth group size fails on CPU — that also happens on `main` with plain triangular `lmax=mmax=17`, so it's a backend limitation rather than anything to do with this change. NCCL handles it fine.

### Smaller items

- `truncation.py`: the new docstring lines use `\\max` / `\\min`, but the docstring is already a raw string (`r"""`), so those become literal double backslashes — and `\\` in LaTeX math is a line break, so it'll render broken. Single `\max` / `\min`, like the surrounding text, should do it.
- `# determine maximum degrees based on triangular truncation` is no longer accurate in 8 places: `sht.py:130,277,410,569` and `distributed/distributed_sht.py:143,334,495,671`.
- Could the Changelog entry mention that `SpectralConv` and its distributed twin are unaffected? They force `lmax = mmax = min(...)` themselves, so nothing changes for the layer users, and that's reassuring to spell out.

### One note on memory

Not a blocker, just something worth being aware of: the distributed forward keeps the full `lmax` in its per-rank Legendre block, `(mmax_local, lmax, nlat_local)`. Decoupling `lmax` lets it run up to `nlat` where it was previously capped near `nlon/2`, so that block can grow by roughly that ratio. Fine as long as it's a deliberate choice by the caller, which the explicit opt-in above would also help make clear.

Thanks again — nice, tightly scoped change, and the tests you added are exactly the right ones.

mwyau:
Hi @azrael417, thank you for the review, this is very comprehensive and offers a different viewpoint from Codex (which I used).

I think `truncation="triangular"` is a great idea. My original request was only to enable different `lmax` and `mmax` and not being intrusive, but it changed the default behavior. If we allow a `truncation` keyword we can keep 100% backward compatibility and also support:

- triangular truncation (default)
- rhomboidal truncation (more common in atmospheric science)
- trapezoidal truncation (less common, the current behavior in this PR when `mmax` < `lmax`)

Reference: [CAM6 Scientific Guide](https://acomstaff.acom.ucar.edu/tilmes/CAM_docs/doc/build/html/cam6_scientific_guide/chapter3.html#spectral-algorithm-overview)

Rhomboidal truncation requires a mask. If you are open to it I can update the PR with the `truncation` parameter and all three variations.

azrael417:
Hello Albert,

thanks for this — rhomboidal/trapezoidal support is something we want, and the implementation is careful: the global `m`/`l` offsets are threaded correctly into the mask in the distributed path, and `test_legendre_blocks` coverage for it is exactly the right place to pin that down.

I'd like to ask for one API change before we merge, and I want to explain the constraint driving it, because it's not visible from the outside. This below is my biggest thing I want to think about and that is the main reason why it takes so long to review this. 

## Context: the grid refactor

We have a branch in flight that replaces the `(nlat, nlon, grid_str)` triple throughout the library with a `GridS2` descriptor object. `truncate_sht` becomes `truncate_sht(grid, lmax, mmax)` and sources its bounds from `grid.max_exact_degree` / `grid.max_azimuthal_order`. That branch also anticipates **reduced Gaussian grids**, HEALPix grids and others where `nlon` varies per latitude — so `mmax` becomes a function of latitude, and the spectral coefficient layout stops being a dense `(lmax, mmax)` rectangle.

That matters here because a truncation expressed as *a mask over a dense rectangle* has nowhere to go on a reduced or unstructured grid. We'll eventually need per-order degree extents (a ragged `[lstart_m, lend_m)` per order) rather than a boolean mask.

**I am not asking you to implement any of that.** The dense-mask internals in this PR are fine and consistent with what's already there — `legpoly` zero-inits and never writes `m > l`, so triangular support is already an implicit zero block in a dense table; your `masked_fill_` is the same idiom on the rhomboid's other edge. What I'd like is only that the *public surface* is one we don't have to deprecate when the refactor lands.

## The request: three numbers, no enum

The `truncation=` string is redundant with the numbers already being passed, and it's what forces `_sht_l_minus_m_max` to reverse-derive `lmax - mmax`. If the `l - m` bandwidth is just a third parameter, all three classical truncations fall out of the values:

```
keep (l, m)  iff  m < mmax  and  m <= l < min(lmax, m + lmmax)
```

- `lmax == mmax`   → triangular
- `lmmax is None`  → trapezoidal
- `lmmax == mmax`  → rhomboidal

This is the classical pentagonal truncation (Sela 1980; NCEP GFS), and it's strictly more general than the three named modes — it also covers e.g. R42 capped at `lmax=64`, which the current PR can't express.

It's also less code than what's here now. It removes:

- `_sht_l_minus_m_max` entirely — `lmmax` is passed straight through
- the mode-string validation, in both `truncate_sht` and `_sht_l_minus_m_max`
- the mode branching in `truncate_sht`
- `self.truncation = truncation` in eight classes, and the `truncation=` threading through the tests
- `test_invalid_truncation` and `test_rhomboidal_requires_explicit_bounds` — with an explicit `lmmax` you can default `lmax` from the grid and still ask for a rhomboid, so the "requires explicit lmax and mmax" restriction disappears

and the user-facing win is that R42 stops requiring mental arithmetic: today it's `lmax=85, mmax=43, truncation="rhomboidal"`, where the 85 has to be derived as 2·42+1.

## Specifics

**1. Name it `lmmax`, and make it non-inclusive.** `kmax` is taken — it's the latitude index bound in `_precompute_legpoly`. `lmmax` parallels `lmax`/`mmax`, and those are both non-inclusive in this codebase, so the bandwidth should be too: `l - m < lmmax`, not `<=`. This changes the constant relative to your current patch, so worth being pedantic about — an off-by-one here is a silently wrong spectral filter rather than a crash.

Under that convention R42 is `lmax=85, mmax=43, lmmax=43`, with the invariant `lmax == mmax + lmmax - 1`.

**2. `lmin`/`mmin` — a naming collision on our side, nothing for you to do.** Worth flagging so it doesn't surprise you later. In the precompute helpers, `mmin`/`lmin` currently mean *the local shard window*, not a truncation — and so, it turns out, do `mmax`/`lmax`: `distributed_sht.py` calls them as `mmax=self.mmax_offset + self.mmax_local, mmin=self.mmax_offset`. So the same four names carry a physical meaning at the module level and a distributed-bookkeeping meaning one layer down. That's our sloppiness, not anything in your patch.

We'll clean it up by renaming the window arguments (`mmin_local`, `mmax_local`, …) so the bare names stay free for truncation semantics — an `lmin`/`mmin` band-pass is something we'd like to be able to express and query eventually, and it fits the same formula above with `mmin <= m < mmax` and `max(lmin, m) <= l`.

The only implication for this PR: no need to add user-facing `lmin`/`mmin` now, and the per-rank restriction is best left derived internally from the global truncation rather than passed in — which is already what you're doing.

**3. Raise instead of clamping when the rhomboid would be reshaped.** `truncate_sht(..., lmax=5, mmax=9, truncation="rhomboidal")` currently clamps to `mmax=5` and yields `l - m <= 0` — a diagonal-only transform — and the test asserts `(5, 5)` as though that were intended. Clamping `mmax` changes the *shape* of the rhomboid, not just its extent, so it should be an error.

**4. Keep the internal representation unified even though the API is flat.** This is the one thing that would save real work later. Rather than three ints scattered across each module, have `truncate_sht` return a single value the modules hold once:

```python
self._trunc = truncate_sht(nlat, nlon, lmax, mmax, lmmax, grid)   # NamedTuple for now

@property
def lmax(self):  return self._trunc.lmax
@property
def mmax(self):  return self._trunc.mmax
@property
def lmmax(self): return self._trunc.lmmax
```

and pass `self._trunc` down to `_precompute_legpoly` / `_precompute_dlegpoly` instead of loose ints. When the grid refactor lands, that NamedTuple gets swapped for a class carrying `l_extents()` / `modes()` / `restrict()` and nothing public moves — no call-site churn, and the distributed classes stop recomputing the bandwidth in four places.

## One piece that's ours, not yours

`truncate_sht` currently forces `lmax = mmax = min(lmax, mmax)`. Making them independent silently changes the meaning of every existing call that passes both, so it needs a deprecation warning for a release. That's a back-compat decision on our users and we should write it — don't feel you need to.

## How would you like to proceed?

Happy either way:

- you take the changes above and we review, or
- if you'd rather not carry the redesign, we can push the changes directly to your branch (append-only, no force-push), or take it over as a maintainer PR with your commits and authorship preserved.

Also — out of curiosity, for the `spharmgrid` backend, do you need rhomboidal for round-tripping existing NCL/Spherepack output, or as a first-class truncation for new work? If it's the former, it's worth us knowing whether the coefficient *layout* needs to match Spherepack's too, not just the retained mode set.

azrael417:
Hello Albert,

sorry for the confusion. Maybe you misunderstood what I meant. Instead of me trying to explain, I created a new branch off of yours: https://github.com/NVIDIA/torch-harmonics/tree/tkurth/sht-truncation-lmmax

Something like this is what we had in mind for the truncation and it decouples the truncation modes from the distributed application. This also goes along with the grid descriptor changes we want to adopt later. Can you please have a look and let me know what you think. Feel free to merge this branch into this here or pick whatever you like but most importantly let me know what you think. Since you are working on a downstream application, it is important for us to understand your potential blockers and requirements, since we would like to enable your application as best as we can.

Best regards
Thorsten

mwyau:
Just rebased. I have an implementation but I am still testing it then I will push to this branch.

PR #233, opened 2026-07-28
Enable the tcgen05 DISCO kernels on SM103 (GB300 / B300)

albertocarpentieri:
## Problem

The Blackwell tensor-core DISCO forward kernel is unreachable on GB300. Its body is gated on `__CUDA_ARCH_FEAT_SM100_ALL`, which nvcc defines only for `-arch=sm_100a`, and `setup.py` only sets `BUILD_KPACKED_SM100` when `TORCH_CUDA_ARCH_LIST` names an arch-conditional `10.0a`. On an SM103 device every DISCO convolution therefore falls back to the CUDA-core CSR kernel, and no build configuration changes this: an `sm_100a` cubin does not load on `sm_103`.

Building for `sm_103a` is not a fix on its own, and fails in a way that is easy to miss. The guard goes undefined, the kernel body compiles to an empty function, and `disco_cuda_fwd_kpacked_sm100` — which never checks the feature macro — allocates its output with `torch::zeros` and launches it. The result is all-zero activations, with a successful launch, no error at any layer, and `_kpacked_supported_on_device()` reporting support because it gates on `major == 10`.

## Change

- Gate the tcgen05 wrappers and kernel body on `DISCO_TCGEN05_SUPPORTED`, defined as `__CUDA_ARCH_FEAT_SM100_ALL || __CUDA_ARCH_FEAT_SM103_ALL`.
- Teach `_torch_cuda_arch_list_has` about `10.3a`, and set `BUILD_KPACKED_SM100` for either Blackwell target. One flag covers both, matching the runtime dispatch on `props.major == 10`.
- Document in the `#else` branch that it yields zeros rather than an error, so the next person adding an architecture validates numerically instead of trusting the build flag.

No API change, and no behavior change on SM90 or SM100. Users on SM103 build with `TORCH_CUDA_ARCH_LIST="10.3a"`; kernel selection at runtime remains automatic.

## Why not the family target

`sm_103f` also assembles tcgen05 and would cover the whole family in one binary, but it defines only `__CUDA_ARCH_FAMILY_SPECIFIC__`, which is equally set for `sm_120f`, where tcgen05 does not exist. Gating on it would be unsound.

The macro behaviour is reproducible with nvcc alone (checked on CUDA 12.9.86):

```bash
cat > probe.cu <<'EOF'
__global__ void k() {
#if defined(__CUDA_ARCH_FEAT_SM100_ALL)
    asm volatile("// MARK_FEAT_SM100_ALL");
#endif
[...]

PR #267, opened 2026-09-21
Add SM103 (GB300 / B300) to wheels build

mwyau:
- #233 added SM103 (GB300 / B300) support but the CI `build_wheels.yml` was not updated.
- Updated CUDA_ARCH_MAP to include 10.3a.


mwyau:
Deleted by mistake, is this still worth to pursue?

